PR #1, opened 2024-05-16
add jupyter notebook

Tianhao-Gu:


README.md
```diff
@@ -1,4 +1,92 @@
...
-environment tailored for executing Spark jobs via Jupyter notebooks.
\ No newline at end of file
+environment tailored for executing Spark jobs via Jupyter notebooks.
+
+## Using `docker-compose.yaml`
+
+To deploy the JupyterHub container and Spark nodes locally, execute the following command:
+
+```bash
+docker-compose up --build
```
MrCreosote, line 11: You can daemonize with `-d`

README.md
```diff
@@ -1,4 +1,92 @@
...
+```
+
+## Test Submitting a Spark Job Locally
+
+### Submitting a Spark Job via spark-test-node
+```bash
+docker exec -it spark-test-node \
+    sh -c '
+	    /opt/bitnami/spark/bin/spark-submit \
+	    --master spark://spark-master:7077 \
```
MrCreosote, line 21: Can you make the spark master an environment variable in docker-compose?

README.md
```diff
@@ -1,4 +1,92 @@
...
+
+### Spark Session/Context Configuration
+
+Ensure to configure `spark.driver.host` for the Spark driver to bind to the Jupyter notebook container's hostname
+
+```python
+spark = SparkSession.builder \
+    .master(os.environ['SPARK_MASTER_URL']) \
+    .appName("TestSparkJob") \
+    .config("spark.driver.host", os.environ['SPARK_DRIVER_HOST']) \
```
MrCreosote, line 68: We should just add this to the config in the image in the entrypoint so the user doesn't have to worry about it

README.md
Tianhao-Gu: 👍

README.md
Tianhao-Gu: I know. But I usually want to see the logs.

MrCreosote: `docker compose logs`

but if you want to leave it as is that's fine

README.md
MrCreosote: :+1:

README.md
Tianhao-Gu: are you referring to make changes to the Spark configuration file? I can do it in my next PR. I think it will involve a lot of testing commit since locally it always works.

README.md
Tianhao-Gu: I know you can view logs that way. I think it's convenient to view it as it's happening. I will just leave it as is for now. It's individual dev's choice I guess.

README.md
MrCreosote: That's what I was thinking, but regardless of the implemenation the behavior should be that we set some environment var in rancher to the notebook hostname and then the user doesn't need to worry about the host name at all, it Just Works

next PR is fine

docker-compose.yaml
```diff
@@ -0,0 +1,62 @@
...
+    image: bitnami/spark:3.5.1
+    container_name: spark-worker-1
+    ports:
+      - "8081:8081"
+    environment:
+      - SPARK_MODE=worker
+      - SPARK_MASTER_URL=spark://spark-master:7077
+      - SPARK_WORKER_CORES=2
+      - SPARK_WORKER_MEMORY=1G
+      - SPARK_WORKER_PORT=8091
```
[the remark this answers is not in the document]

docker-compose.yaml
```diff
@@ -0,0 +1,62 @@
...
+    environment:
+      - SPARK_MASTER_URL=spark://spark-master:7077
+
+  notebook:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: spark-notebook
+    ports:
+      - "4040:4040"
```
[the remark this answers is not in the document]
Tianhao-Gu: removed. I must test something before and I forget to remove it.

README.md
```diff
@@ -1,4 +1,92 @@
...
+    .getOrCreate()
+```
+```python
+conf = SparkConf(). \
+    setMaster( os.environ['SPARK_MASTER_URL']). \
+    setAppName("TestSparkJob"). \
+    set("spark.driver.host", os.environ['SPARK_DRIVER_HOST'])
+sc = SparkContext(conf=conf)
```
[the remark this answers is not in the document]
Tianhao-Gu: No, you don't need to set up both. SparkContext has more control over spark but more complicated to config. Some people have a preference for either option. So I just give 2 examples.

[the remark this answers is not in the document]
MrCreosote: Maybe put "OR" between them?

docker-compose.yaml
```diff
@@ -0,0 +1,62 @@
+version: '3'
+
+# This docker-compose is for developer convenience, not for running in production.
+
+services:
+
+  spark-master:
+    image: bitnami/spark:3.5.1
+    container_name: spark-master
```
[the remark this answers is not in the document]
Tianhao-Gu: Otherwise the container name is <hostname>_1. I just don't like the suffix when I need to `docker exec -it`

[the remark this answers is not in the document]
MrCreosote: :+1:

README.md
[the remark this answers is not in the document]
Tianhao-Gu: 👍

docker-compose.yaml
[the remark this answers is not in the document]

docker-compose.yaml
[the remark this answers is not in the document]
MrCreosote: :

README.md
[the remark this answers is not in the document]
MrCreosote: :+1:

Dockerfile
```diff
@@ -1,4 +1,23 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Jupyterlab and other python dependencies
+RUN pip3 install jupyterlab==4.2.0 pyspark==3.5.1
+
+COPY scripts/entrypoint.sh /opt/
+RUN chmod a+x /opt/entrypoint.sh
+
+# Switch back to the original user
+#USER ${ORI_USER}
```
MrCreosote, line 21: is this supposed to be commented out now?

Tianhao-Gu: oh oh. I was testing things and did `add .`

Dockerfile
```diff
@@ -1,4 +1,23 @@
 FROM bitnami/spark:3.5.1
 
+RUN export ORI_USER=$(id -u)
+# Switch to root to install packages
+USER root
 
-ENTRYPOINT ["sleep 10s"]
+ENV PYTHON_VER=python3.11
```
MrCreosote, line 7: Now I think about it I'm pretty sure bitnami installs python in the image. Is there a reason you're not using that python, assuming I'm right? If the python version here doesn't match what's in the cluster the job won't run

Tianhao-Gu: yea. I can use python from the image. But I remember you want to control which python version we are using. I have no strong option on that. I am fine either way.

MrCreosote: In the prior repo the version was set for both the cluster and the test image (assuming you were using the same image). Here the python version could be different for the cluster and the notebook which would break things

Tianhao-Gu: 👍 Switched to use image python. Both base image has python installed. I don't see the differences.

MrCreosote: > Both base image has python installed.

What do you mean by both base images?

>  I don't see the differences.

Not sure what you mean here either

Tianhao-Gu: both `bitnami/spark` and the image you used previously has python. I don't see why we spent a lot of time trying to install python before. We could just use the python form the image before.

MrCreosote: The old image had 3.10. If it had 3.11 I wouldn't have bothered

MrCreosote (approved):

Dockerfile

PR #3, opened 2024-05-20
add vol mount

Tianhao-Gu:


docker-compose.yaml
```diff
@@ -63,4 +63,6 @@ services:
     environment:
       - NOTEBOOK_PORT=4041
       - SPARK_MASTER_URL=spark://spark-master:7077
-      - SPARK_DRIVER_HOST=spark-notebook
\ No newline at end of file
+      - SPARK_DRIVER_HOST=spark-notebook
+    volumes:
+      - ./cdm_shared_workspace:/cdm_shared_workspace
```
MrCreosote, line 68: Where are you mounting from in rancher?

docker-compose.yaml
MrCreosote, line 1: At some point we'll probably have to think about how different users would interact with this but that's much later

docker-compose.yaml
Tianhao-Gu: in Rancher I just do `/cdm_shared_workspace:/cdm_shared_workspace`.

MrCreosote: Where does the data live though? In the VM the notebook is running on?

Tianhao-Gu: I think so. Double confirming with devops team in slack.

MrCreosote:
is this still relevant?

Tianhao-Gu:
> is this still relevant?

no. Just closed.

PR #4, opened 2024-05-20
add spark helper

Tianhao-Gu:


README.md
```diff
@@ -63,23 +63,37 @@ When running Spark in the Jupyter notebook container, the default `spark.driver.
 the hostname (`SPARK_DRIVER_HOST`) of the container. 
 In addition, the environment variable `SPARK_MASTER_URL` should also be configured.
 
+#### Using Predefined SparkSession from `spark.utils.get_spark_session` method
+```python
+from spark.utils import get_spark_session
```
MrCreosote, line 68: Will `spark` not conflict with other package names?

src/__init__.py
MrCreosote, line 1: I don't think `__init__` files are necessary any more

README.md
Tianhao-Gu: I don't think so. But just in case changed the repo name to `common`.

src/__init__.py
Tianhao-Gu: 👍 removed.

README.md
MrCreosote: https://pypi.org/project/common/ seems dead.

That being said, https://pypi.org/project/spark/ seems super dead

MrCreosote: So spark seems safe and is probably a better name, sorry about that

Tianhao-Gu: 👍

PR #5, opened 2024-05-20
add tests

Tianhao-Gu:


MrCreosote (approved):
Not sure we really need tests at this stage of the game but doesn't hurt

PR #8, opened 2024-05-20
address dependabot upgrades

Tianhao-Gu:


.github/workflows/test.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       run: PYTHONPATH=. pytest --cov=src --cov-report=xml test
 
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
-        fail_ci_if_error: true
\ No newline at end of file
+        fail_ci_if_error: false
```
MrCreosote, line 50: Do you really want to do this? It's true everywhere else, and not having coverage in general is a good reason to fail ci

Tianhao-Gu: 👍 turned it back on. Token wasn't working before. But now all good.

MrCreosote (approved):

PR #9, opened 2024-05-22
add delta spark packages to support delta lake

Tianhao-Gu:


Dockerfile
```diff
@@ -6,9 +6,23 @@ USER root
...
+RUN curl -O https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-aws/${HADOOP_AWS_VER}/hadoop-aws-${HADOOP_AWS_VER}.jar \
+    && mv hadoop-aws-${HADOOP_AWS_VER}.jar /opt/bitnami/spark/jars
+
+# NOTE: ensure Delta Spark jars matche python pip delta-spark version specified in the Pipfile
+ENV DELTA_SPARK_VER=3.2.0
+ENV SCALA_VER=2.12
+RUN curl -O https://repo1.maven.org/maven2/io/delta/delta-spark_${SCALA_VER}/${DELTA_SPARK_VER}/delta-spark_${SCALA_VER}-${DELTA_SPARK_VER}.jar \
+    && mv delta-spark_${SCALA_VER}-${DELTA_SPARK_VER}.jar /opt/bitnami/spark/jars
+
+Run curl -O https://repo1.maven.org/maven2/io/delta/delta-storage/${DELTA_SPARK_VER}/delta-storage-${DELTA_SPARK_VER}.jar \
```
MrCreosote, line 23:
```suggestion
RUN curl -O https://repo1.maven.org/maven2/io/delta/delta-storage/${DELTA_SPARK_VER}/delta-storage-${DELTA_SPARK_VER}.jar \
```

Dockerfile
```diff
@@ -6,9 +6,23 @@ USER root
...
+    && mv delta-storage-${DELTA_SPARK_VER}.jar /opt/bitnami/spark/jars
```
MrCreosote, line 24: None of these have dependencies? 

Even if not, I'd probably do this with Gradle. Setting up a dependencies only build file is pretty easy

docker-compose.yaml
```diff
@@ -51,10 +53,48 @@ services:
...
+      - "9000:9000"
+      # MinIO Console is available at http://localhost:9001
+      - "9001:9001"
+    environment:
+      MINIO_ROOT_USER: minio
+      MINIO_ROOT_PASSWORD: minio123
+      MINIO_ACCESS_KEY: minio
+      MINIO_SECRET_KEY: minio123
+    healthcheck:
+      test: timeout 5s bash -c ':> /dev/tcp/127.0.0.1/9000' || exit 1
```
MrCreosote, line 71: This line could use some explanation

docker-compose.yaml
```diff
@@ -51,10 +53,48 @@ services:
...
+      interval: 1s
+      timeout: 10s
+      retries: 5
+    # Note there is no bucket by default
+    command: server /data --console-address ":9001"
```
MrCreosote, line 76: The console isn't on by default?

docker-compose.yaml
```diff
@@ -51,10 +53,48 @@ services:
...
+      else
+        echo 'bucket delta-lake already exists'
+      fi
+      "
+
   notebook:
     build:
       context: .
       dockerfile: Dockerfile
+    entrypoint: /opt/scripts/notebook_entrypoint.sh
```
MrCreosote, line 97: This means we're going to have to configure the entrypoint in rancher, which devops is generally trying to avoid

docker-compose.yaml
```diff
@@ -51,10 +53,48 @@ services:
...
+      MINIO_SECRET_KEY: minio123
+    healthcheck:
+      test: timeout 5s bash -c ':> /dev/tcp/127.0.0.1/9000' || exit 1
+      interval: 1s
+      timeout: 10s
+      retries: 5
+    # Note there is no bucket by default
+    command: server /data --console-address ":9001"
+
+  minio-create-bucket:
```
MrCreosote, line 78: Seems like nothing depends on this?

docker-compose.yaml
```diff
@@ -63,4 +103,9 @@ services:
...
       - NOTEBOOK_PORT=4041
       - SPARK_MASTER_URL=spark://spark-master:7077
-      - SPARK_DRIVER_HOST=spark-notebook
\ No newline at end of file
+      - SPARK_DRIVER_HOST=spark-notebook
+      - MINIO_URL=http://minio:9000
+      - MINIO_ACCESS_KEY=minio
+      - MINIO_SECRET_KEY=minio123
+    volumes:
+      - ./cdm_shared_workspace:/cdm_shared_workspace
```
MrCreosote, line 111: I thought from slack we didn't want this mount?

src/spark/utils.py
```diff
@@ -19,14 +22,39 @@ def get_spark_session(app_name: str, local: bool = False) -> SparkSession:
...
+                ),
+                ("spark.app.name", app_name),
+                ("spark.hadoop.fs.s3a.endpoint", os.environ.get("MINIO_URL")),
+                ("spark.hadoop.fs.s3a.access.key", os.environ.get("MINIO_ACCESS_KEY")),
+                ("spark.hadoop.fs.s3a.secret.key", os.environ.get("MINIO_SECRET_KEY")),
+                ("spark.jars", jars),
+                ("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension"),
+                ("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog"),
+                ("spark.hadoop.fs.s3a.path.style.access", "true"),
+                ("spark.hadoop.fs.s3a.connection.ssl.enabled", "false"),
```
MrCreosote, line 45: This should probably be a parameter so we don't accidentally make everything insecure later

Dockerfile

Dockerfile
Tianhao-Gu: I honestly have barely no knowledge how Gradle works. Maybe I can learn it at a later time unless you think it's necessary to get it done now.

docker-compose.yaml
Tianhao-Gu: 👍

Dockerfile
MrCreosote: It's not hard, but a TODO is fine

docker-compose.yaml
Tianhao-Gu: yea. I configured this in Rancher. I cannot think of other ways to use this as both master and notebook.

docker-compose.yaml
Tianhao-Gu: I don't think we need it. Just local testing environment.

docker-compose.yaml
Tianhao-Gu: we don't want this for Rancher. But this is just for local development.

src/spark/utils.py
Tianhao-Gu: I just removed this and leave it as default. I think it doesn't matter here.

docker-compose.yaml
MrCreosote: :+1:

docker-compose.yaml
MrCreosote: Have an entrypoint that takes an env var that selects the correct mode

docker-compose.yaml
MrCreosote: You mean you're going to delete it?

docker-compose.yaml
MrCreosote: ok, maybe to avoid confusion we should replicate the rancher config here

Dockerfile

docker-compose.yaml
MrCreosote: I'm still confused about the `:>` part. Also tcp != http

docker-compose.yaml
Tianhao-Gu: We don't really know what the base image does for different mode.

docker-compose.yaml
Tianhao-Gu: I mean i don't think we need anything to depends on it.

src/spark/utils.py
MrCreosote: :+1:

docker-compose.yaml
MrCreosote: I'm not sure what you mean, the base image is pretty well documented

docker-compose.yaml
MrCreosote: Seems like the notebook should if it's expecting a bucket to exist

Dockerfile

docker-compose.yaml
Tianhao-Gu: maybe I will just add the reference.

docker-compose.yaml
Tianhao-Gu: 👍

docker-compose.yaml
Tianhao-Gu: unnecessary change but 👍 . Rancher settings can change.

docker-compose.yaml
Tianhao-Gu: 👍 but again, not worth it and might cause problem if base image is updated to use a different entrypoint script. Way easier to maintain by just giving Rancher the entrypoint script.

MrCreosote: Not for the people setting up the images. If the base image entrypoint changes we just change ours

docker-compose.yaml
MrCreosote: :+1:

docker-compose.yaml
MrCreosote: :+1:

docker-compose.yaml

docker-compose.yaml
MrCreosote: If there's an ENTRYPOINT there's no need for CMD, is there?

MrCreosote: Actually I think the current setup will do something weird, as the contents of CMD will be provided as arguments to ENTRYPOINT IIUC

Tianhao-Gu: I tried to put `/opt/bitnami/scripts/spark/run.sh` at the end of main entrypoint.sh file. The master services failed to start.

Tianhao-Gu: so in `/opt/bitnami/scripts/spark/entrypoint.sh`

```
if [[ "$1" = "/opt/bitnami/scripts/spark/run.sh" ]]; then
    info "** Starting Spark setup **"
    /opt/bitnami/scripts/spark/setup.sh
    info "** Spark setup finished! **"
fi
```

Tianhao-Gu: removed `run.sh` to the entypoint as arguments and added notes to explain it.

MrCreosote (approved):
LGTM

PR #11, opened 2024-05-23
using hive for metastore

Tianhao-Gu:


src/spark/utils.py
```diff
@@ -3,57 +3,94 @@
...
-    return SparkSession.builder.config(conf=spark_conf).getOrCreate()
+        # Just to include the necessary jars for Delta Lake
+        jar_names = [f"delta-spark_{SCALA_VER}-{DELTA_SPARK_VER}.jar",
+                     f"hadoop-aws-{HADOOP_AWS_VER}.jar"]
+        jars_str = _get_jars(jar_names)
+        delta_conf = _get_delta_lake_conf(jars_str)
+        for key, value in delta_conf.items():
+            spark_conf.set(key, value)
+
+    return SparkSession.builder.config(conf=spark_conf).enableHiveSupport().getOrCreate()
```
MrCreosote, line 96: I don't know anything about hive, really - why is it needed here? Doesn't DeltaLake take care of the metadata?

Tianhao-Gu: with this approach, we can establish a permanent view that everyone can query without rebuilding/reloading the table. I removed `enableHiveSupport()` from general spark session builder and only enable this via conf for delta lake.

MrCreosote: So the delta lake SW doesn't maintain the table metadata anywhere other than local memory?

Where are the metadata persisted?

Tianhao-Gu: Yea, in memory by default. The metadata is persisted in the metastore_db folder which is created by default under /cdm_shared_folder and mounted by the rancher settings.

MrCreosote: So anyone wanting to use the shared tables needs the same volume mount for their notebook container? What happens if two people try to create conflicting tables with the same name at the same time?

Tianhao-Gu: I believe table is locked for concurrent access. I don’t understand the first question. Does that matter?

MrCreosote: The table metadata is stored in a file on the volume mount, right? So If you're working in notebook container A and create a table, that metadata gets stored in the volume mount, which is actually some real place outside the container. If I'm working in notebook container B, then I need that same file mounted into my container to see the table you made.

Will the concurrent locks work though the different containers? E.g. if I lock  the files in container B will they be locked for you in container A?

Tianhao-Gu: Yea the metadata is stored in the shared volume. If you are looking to read metadata the container need to be mounted the same volume which is exactly our current setup. The lock is on the metadata file so it’s locked for all containers.

MrCreosote: The reason I'm asking is that for some flavors of NFS os level file locks don't actually work and allow writing to the file on other machines while it's locked on the first machine. With how quirky docker can be I'm wondering if it has the same sort of problems

Tianhao-Gu: yea I notice that behavior. In Jupiter notebook I gave some instructions of how to delete the DB locks manually in the event if accessing issues.

MrCreosote: What's the lock mechanism?

Tianhao-Gu: There is a .lck file. Just delete that file.

MrCreosote: Ok, that seems like that should work regardless of docker

MrCreosote: Can you check and see if there's anyway for the tables to be stored in minio so other systems that can talk deltalake can access them?

MrCreosote (approved):

src/spark/utils.py
Tianhao-Gu: yea. I will look into that.

PR #15, opened 2024-05-31
set spark session timeout

Tianhao-Gu:
auto stop spark session to prevent idle session from consuming all resources 
Screenshot 2024-05-31 at 1 26 48 PM

### DIFF
--- a/src/spark/utils.py
+++ b/src/spark/utils.py
@@ -9,7 +9,6 @@ JAR_DIR = '/opt/bitnami/spark/jars'
 HADOOP_AWS_VER = os.getenv('HADOOP_AWS_VER')
 DELTA_SPARK_VER = os.getenv('DELTA_SPARK_VER')
 SCALA_VER = os.getenv('SCALA_VER')
-SPARK_TIMEOUT_SECONDS = os.getenv('SPARK_TIMEOUT_SECONDS', 4 * 60 * 60)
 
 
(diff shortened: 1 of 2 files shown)


MrCreosote:
So your session is killed 24h after you created it, no matter what?

Tianhao-Gu:
> So your session is killed 24h after you created it, no matter what?

after 4 hours. I don't anticipate anyone will use spark session longer than that. I noticed people don't stop spark session properly. We need to enforce closing sessions. As the screenshot shows, currently there is an open spark session running for 24 hours taking all resources.

MrCreosote:
> after 4 hours. 

oops

MrCreosote (approved):
This seems ok as a temporary measure, but I can see cases where people might be doing complicated computations / work and have legit sessions take > 4h. We probably need to figure out a better solution long term - maybe bring this up in the checkpoint next week?

Tianhao-Gu:
> This seems ok as a temporary measure, but I can see cases where people might be doing complicated computations / work and have legit sessions take > 4h. We probably need to figure out a better solution long term - maybe bring this up in the checkpoint next week?

how about making timeout_seconds as an argument to the function. So that if people want a longer session, they can specify it during func calls.

MrCreosote:
That also seems fine for now, but long term I'm not sure it'll work, hence discussion next week

PR #22, opened 2024-06-04
auto assigning spark app name

Tianhao-Gu:


src/spark/utils.py
```diff
@@ -72,20 +73,23 @@ def get_base_spark_conf(app_name: str) -> SparkConf:
...
+    :param app_name: The name of the application. If not provided, a default name will be generated.
+    :param local: Whether to run the spark session locally or not. Default is False.
+    :param delta_lake: Build the spark session with Delta Lake support. Default is True.
+    :param timeout_sec: The timeout in seconds to stop the Spark session forcefully. Default is 4 hours.
 
     :return: A SparkSession object
     """
+    if not app_name:
+        app_name = f"kbase_spark_session_{datetime.now().strftime('%Y%m%d%H%M%S')}"
+
```
MrCreosote, line 92: If there is an app name should it be suffixed by the timestamp to avoid name collisions?

MrCreosote (approved):

src/spark/utils.py
Tianhao-Gu: If an app_name is provided by the user, they might want to use existing session to avoid creating new ones. I think we probably shouldn't add suffix automatically.

Tianhao-Gu: Yea. There can always be multiple apps(sessions) per driver. We just turned the setting so that sessions can run concurrently. Otherwise, the queue is FIFO by default.

PR #23, opened 2024-06-06
config executor cores

Tianhao-Gu:
With the changes, we can control how many cores to use for the spark session.

### DIFF
--- a/src/spark/utils.py
+++ b/src/spark/utils.py
@@ -10,7 +10,9 @@ JAR_DIR = '/opt/bitnami/spark/jars'
 HADOOP_AWS_VER = os.getenv('HADOOP_AWS_VER')
 DELTA_SPARK_VER = os.getenv('DELTA_SPARK_VER')
 SCALA_VER = os.getenv('SCALA_VER')
-DEFAULT_EXECUTOR_CORES = 1  # the default number of CPU cores that each Spark executor will use
+# the default number of CPU cores that each Spark executor will use
+# If not specified, Spark will typically use all available cores on the worker nodes
+DEFAULT_EXECUTOR_CORES = 1
 
 
 def _get_jars(jar_names: list) -> str:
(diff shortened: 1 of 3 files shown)


src/spark/utils.py
```diff
@@ -10,6 +10,7 @@
 HADOOP_AWS_VER = os.getenv('HADOOP_AWS_VER')
 DELTA_SPARK_VER = os.getenv('DELTA_SPARK_VER')
 SCALA_VER = os.getenv('SCALA_VER')
+DEFAULT_EXECUTOR_CORES = 1  # the default number of CPU cores that each Spark executor will use
```
MrCreosote, line 13: Without this argument how many cores does each executor use? Can you document that here?

Tianhao-Gu: added comments. I noticed Spark will take all cores on the worker if we don't set this.

src/spark/utils.py
```diff
@@ -29,11 +32,14 @@ def _get_jars(jar_names: list) -> str:
...
 
-def _get_delta_lake_conf(jars_str: str) -> dict:
+def _get_delta_lake_conf(
+        jars_str: str,
+        executor_cores: int) -> dict:
     """
     Helper function to get Delta Lake specific Spark configuration.
 
     :param jars_str: A comma-separated string of JAR file paths
+    :param executor_cores: The number of CPU cores that each Spark executor will use
```
MrCreosote, line 42: So this only applies if Delta lake is true and the user doesn't call the get base conf method. It seems like it should always apply

Tianhao-Gu: 👍

src/spark/utils.py
```diff
@@ -58,32 +63,39 @@ def _stop_spark_session(spark):
     spark.stop()
 
 
-def get_base_spark_conf(app_name: str) -> SparkConf:
+def get_base_spark_conf(
```
MrCreosote, line 66: Should this be a private method? If not executor_cores should probably be a default arg

Tianhao-Gu: 👍 i think it makes sense to make it a private method.

MrCreosote (approved):

PR #25, opened 2024-06-06
add minio client helper

Tianhao-Gu:


src/minio_utils/minio_utils.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def get_minio_client() -> Minio:
+    """
+    Helper function to get the Minio client.
+
+    :return: A Minio client object
+    """
+    return Minio(
+        os.environ['MINIO_URL'].replace("http://", ""),
```
MrCreosote, line 17: here I'm wondering why we're just not excluding `http://` in the env var in rancher. Is there a reason for that?

Tianhao-Gu: tested removing `http://` from docker-compose file. When I tried to read delta files, it took 10 minutes without returning anything. I am guessing it's needed for spark custom s3 endpoint config.

MrCreosote (approved):

PR #26, opened 2024-06-06
add helper func to read csv file

Tianhao-Gu:


src/spark/utils.py
```diff
@@ -122,3 +126,52 @@ def get_spark_session(
...
+        path: str,
+        header: bool = True,
+        sep: str = None,
+        **kwargs
+) -> DataFrame:
+    """
+    Read a file in CSV format from minIO into a Spark DataFrame.
+
+    :param spark: The Spark session.
+    :param path: The minIO path to the CSV file.
```
MrCreosote, line 159: An example would help here given the parsing below

src/spark/utils.py
```diff
@@ -122,3 +126,52 @@ def get_spark_session(
...
+    :return: A DataFrame.
+    """
+
+    if not sep:
+        client = get_minio_client()
+        bucket, key = path.replace("s3a://", "").split("/", 1)
+        obj = client.get_object(bucket, key)
+        sample = obj.read(1024).decode()
+        sep = _detect_delimiter(sample)
+        print(f"Detected delimiter: {sep}")
```
MrCreosote, line 173: This can return None. What happens in that case?

src/spark/utils.py
Tianhao-Gu: 👍

src/spark/utils.py
Tianhao-Gu: spark will use the default delimiter `,`. Added comments for this behavior as well.

Tianhao-Gu: maybe we should just terminate if we cannot decide and having users to manually put the delimiter themselves?

MrCreosote: I'd guess if the sniffer can't figure it out it's probably not `,`, so terminating seems safer

MrCreosote: Unless 1024 bytes isn't enough to figure it out - first column is really big for example

src/spark/utils.py
MrCreosote: :+1:

src/spark/utils.py
```diff
@@ -122,3 +126,52 @@ def get_spark_session(
...
+
+    :return: The detected delimiter or raise error if it cannot be detected
+    """
+
+    try:
+        sniffer = csv.Sniffer()
+        dialect = sniffer.sniff(sample)
+        return dialect.delimiter
+    except Exception as e:
+        raise ValueError(f"Delimiter could not be detected: {e}. Please provide the delimiter explicitly.")
```
MrCreosote, line 145:
```suggestion
        raise ValueError(
            f"Delimiter could not be detected: {e}. Please provide the delimiter explicitly.") from e
```

src/spark/utils.py
```diff
@@ -122,3 +126,52 @@ def get_spark_session(
...
+    :param kwargs: Additional arguments to pass to spark.read.csv.
+
+    :return: A DataFrame.
+    """
+
+    if not sep:
+        client = get_minio_client()
+        bucket, key = path.replace("s3a://", "").split("/", 1)
+        obj = client.get_object(bucket, key)
+        sample = obj.read(1024).decode()
```
MrCreosote, line 171:
```suggestion
        sample = obj.read(8192).decode()
```

src/spark/utils.py
MrCreosote: :+1:

MrCreosote (approved):
assuming tests pass

PR #27, opened 2024-06-07
using postgres as hive metastore

Tianhao-Gu:
Using PostgreSQL instead of the default Derby DB will resolve the concurrency issues

### DIFF
--- a/config/hive-site-template.xml
+++ b/config/hive-site-template.xml
@@ -20,10 +20,10 @@
   </property>
 
   <!-- Configuration to automatically create the necessary tables in the database schema if they don't exist -->
-  <!-- Hive metastore fails to start if this property is not set-->
+  <!-- The Hive metastore may fail to start if this property is unset and tables have not been created previously-->
   <property>
     <name>datanucleus.schema.autoCreateTables</name>
-    <value>true</value>
+    <value>{{DATANUCLEUS_AUTO_CREATE_TABLES}}</value>
   </property>
 
   <!-- Disable schema verification in the Hive metastore -->
--- a/config/hive-site-template.xml
+++ b/config/hive-site-template.xml
@@ -1,0 +1,47 @@
+<configuration>
+  <property>
+    <name>javax.jdo.option.ConnectionURL</name>
+    <value>jdbc:postgresql://{{POSTGRES_URL}}/{{POSTGRES_DB}}</value>
+  </property>
+
+  <!-- JDBC driver class name for PostgreSQL -->
+  <!--  Ensure that PostgreSQL JDBC driver jars are included via Gradle -->
+  <property>
+    <name>javax.jdo.option.ConnectionDriverName</name>
+    <value>org.postgresql.Driver</value>
+  </property>
+  <property>
+    <name>javax.jdo.option.ConnectionUserName</name>
+    <value>{{POSTGRES_USER}}</value>
+  </property>
+  <property>
+    <name>javax.jdo.option.ConnectionPassword</name>
+    <value>{{POSTGRES_PASSWORD}}</value>
+  </property>
+
+  <!-- Configuration to automatically create the necessary tables in the database schema if they don't exist -->
+  <!-- Hive metastore fails to start if this property is not set-->
+  <property>
+    <name>datanucleus.schema.autoCreateTables</name>
+    <value>true</value>
+  </property>
+
+  <!-- Disable schema verification in the Hive metastore -->
+  <!-- Hive metastore fails to start if this property is not set-->
+  <property>
+    <name>hive.metastore.schema.verification</name>
+    <value>false</value>
+  </property>
+
+  <!-- Directory location for the Hive warehouse where table data is stored -->
+  <property>
+    <name>hive.metastore.warehouse.dir</name>
+    <value>/cdm_shared_workspace/hive_metastore</value>
+  </property>
+
+  <!-- Enable support for concurrency in Hive, allowing multiple users to access and modify the data simultaneously -->
+  <property>
+    <name>hive.support.concurrency</name>
+    <value>true</value>
+  </property>
+</configuration>
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -100,11 +100,11 @@ services:
       fi
       "
 
-  notebook:
+  dev_notebook:
     build:
       context: .
       dockerfile: Dockerfile
-    container_name: spark-notebook
+    container_name: spark-dev-notebook
     ports:
       - "4041:4041"
     depends_on:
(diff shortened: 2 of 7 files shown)


config/hive-site-template.xml
MrCreosote, line 1: I assume this was faster than getting a hive metastore set up?

config/hive-site-template.xml
```diff
@@ -0,0 +1,47 @@
...
+  <property>
+    <name>javax.jdo.option.ConnectionPassword</name>
+    <value>{{POSTGRES_PASSWORD}}</value>
+  </property>
+
+  <!-- Configuration to automatically create the necessary tables in the database schema if they don't exist -->
+  <!-- Hive metastore fails to start if this property is not set-->
+  <property>
+    <name>datanucleus.schema.autoCreateTables</name>
+    <value>true</value>
```
MrCreosote, line 26: This means the notebook container has to have create table perms on the postgres instance. It seems like it'd be better if the tables were created beforehand and the container only has write to those tables

config/hive-site-template.xml
MrCreosote, line 1: Did you test that this fixes the concurrency issue or are you assuming based on documentation

MrCreosote, line 1: I don't suppose there's an alternative to xml for this config

docker-compose.yaml
```diff
@@ -14,9 +14,18 @@ services:
     environment:
       - SPARK_MODE=master
       - SPARK_MASTER_WEBUI_PORT=8090
+      - MAX_EXECUTORS=4
+      - POSTGRES_USER=hive
+      - POSTGRES_PASSWORD=hivepassword
+      - POSTGRES_DB=hive
+      - POSTGRES_URL=postgres:5432
```
MrCreosote, line 21: Does the master really need the hive creds? I get the workers needing them

docker-compose.yaml
```diff
@@ -14,9 +14,18 @@ services:
...
+    volumes:
+      - ./cdr/cdm/jupyter:/cdm_shared_workspace
```
MrCreosote, line 23: Why do the master and workers need the shard space now?

docker-compose.yaml
```diff
@@ -103,6 +118,29 @@ services:
...
+    ports:
+      - "5432:5432"
+    environment:
+      - POSTGRES_USER=hive
+      - POSTGRES_PASSWORD=hivepassword
+      - POSTGRES_DB=hive
     volumes:
-      - ./cdr/cdm/jupyter:/cdm_shared_workspace
\ No newline at end of file
+      - ./cdr/cdm/jupyter/cdm-postgres:/var/lib/postgresql/data
```
MrCreosote, line 146: Whty does the postgres data need to be shared?

scripts/setup.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
MrCreosote, line 1: is his all the same as from notebook_entrypoint other than the hive stuff at the end?

config/hive-site-template.xml
Tianhao-Gu: a standalone hive metastore? I haven't looked into that yet. I don't know a lot about it. Maybe as next step since starrocks also needs it.

config/hive-site-template.xml
Tianhao-Gu: I don't think any of us fully understand which tables to create. It's probably safer to let it create what it needs.

config/hive-site-template.xml
Tianhao-Gu: it's working locally without any issues.

Tianhao-Gu: yea. As far as I understand it.

docker-compose.yaml
Tianhao-Gu: As I read it, Hive config needs to be present in master, workers as well as driver nodes. That's why I put it there.

docker-compose.yaml
Tianhao-Gu: This is another thing I might not fully understand. But as far as I read it, hive will write to `warehouse.dir`. I am not sure which service is writing to it. But even though, master and workers now have the shared space, the `warehouse.dir` is still empty. It doesn't prevent users to read/query tables. I am guess this setting won't hurt.

docker-compose.yaml
Tianhao-Gu: it doesn't need to. But I think it's convenient if we want to config postgresql on the fly if needed. Like what we did previously to manually delete lock files. Once system is settle, I am planning to mount it differently on rancher.

scripts/setup.sh
Tianhao-Gu: yes. exactly copied over except for the last few line regarding hive-site.xml

docker-compose.yaml
Tianhao-Gu: I am also planning to build a user_notebook and a dev_notebook and mount differently so that users won't accidentally corrupt files.

config/hive-site-template.xml
MrCreosote: Fair enough

config/hive-site-template.xml
MrCreosote: Can we run a notebook with create table privs, have hive create the tables, and then restart the notebook with just write? Or have a separate notebook with just write?

config/hive-site-template.xml
MrCreosote: working with concurrent requests or just working?

MrCreosote: :+1:

docker-compose.yaml
MrCreosote: :+1:

docker-compose.yaml
MrCreosote: Maybe we should have a ticket to look into this later?

docker-compose.yaml
MrCreosote: ok, maybe add a comment about that? If we're running postgres separately in rancher its data folders probably shouldn't be accessible in the notebook

scripts/setup.sh
MrCreosote: :+1:

MrCreosote:
I assume we're going to need to reload data again once this is up in rancher yeah?

config/hive-site-template.xml
Tianhao-Gu: With concurrent requests.

Tianhao-Gu:
> I assume we're going to need to reload data again once this is up in rancher yeah?

Yea. Unfortunately.

config/hive-site-template.xml
Tianhao-Gu: I guess yea we can introduce a env variable to control this behavior. But what if a table needs to be created in the middle of operations? I guess if we are creating a DB separately just for hive, would that be fine? What’s bad about giving hive the privilege to create what it needs?

config/hive-site-template.xml
MrCreosote: :+1:

config/hive-site-template.xml
MrCreosote: If hive can create tables in the db, then so can anyone using the notebook container. Not sure what other privs the creds we'll give to the container provide as well.

Basically the same idea as giving reduced creds for minio to reduce the chance of things getting borked up

Tianhao-Gu: 👍 added `USAGE_MODE` to toggle settings accordingly.

docker-compose.yaml

docker-compose.yaml
Tianhao-Gu: yea. created an issue https://github.com/kbase/cdm-jupyterhub/issues/28. I anticipate that once we set up a standalone Hive service, this issue will no longer be relevant.

Tianhao-Gu:
Tested concurrent querying of the table from both the development and user notebooks, and it works fine.

docker-compose.yaml
MrCreosote: :+1:

docker-compose.yaml
MrCreosote: :

config/hive-site-template.xml
MrCreosote: :+1:

docker-compose.yaml
```diff
@@ -98,11 +113,61 @@ services:
...
+      - NOTEBOOK_PORT=4042
+      - SPARK_MASTER_URL=spark://spark-master:7077
+      - SPARK_DRIVER_HOST=spark-user-notebook
+      - MINIO_URL=http://minio:9002
+      - MINIO_ACCESS_KEY=minio  # TODO: create minIO user and policy for read-only access
+      - MINIO_SECRET_KEY=minio123
+      - SPARK_MODE=notebook
+      - MAX_EXECUTORS=4
+      - POSTGRES_USER=hive
+      - POSTGRES_PASSWORD=hivepassword
```
MrCreosote, line 150:
```suggestion
      # TODO: create postgres user w/ only write access to the hive tables
      - POSTGRES_USER=hive
      - POSTGRES_PASSWORD=hivepassword
```

scripts/setup.sh
MrCreosote, line 1: this is getting big enough we might want to consider porting it to python

MrCreosote (approved):

scripts/setup.sh
Tianhao-Gu: 👍

PR #30, opened 2024-06-11
add SparkMonitor and itables extension

Tianhao-Gu:
SparkMonitor extension:
Screenshot 2024-06-10 at 1 56 37 PM

iTables extension:
Screenshot 2024-06-10 at 1 56 27 PM


src/notebook_utils/notebook_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def display_df(
+        df: DataFrame,
+        layout: dict = None,
+        buttons: list = None,
+        length_menu: list = None
+) -> None:
+    """
+    Display a pandas DataFrame using itables.
```
MrCreosote, line 13: is there a way to do something similar for spark dfs?

src/notebook_utils/notebook_utils.py
```diff
@@ -0,0 +1,41 @@
...
+    iTables project page: https://github.com/mwouts/itables
+
+    :param df: a pandas DataFrame
+    :param layout: layout options, refer to https://datatables.net/reference/option/layout
+    :param buttons: buttons options, options refer to https://datatables.net/reference/button/
+    :param length_menu: length menu options
+    :return:
+    """
+    # initialize itables for the notebook
+    init_notebook_mode(all_interactive=False)
```
MrCreosote, line 23: Is this ok to run many times? Seems like it's something you'd run once

src/notebook_utils/notebook_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        "bottom2Start": "info"
+    }
+    default_buttons = ["csvHtml5", "excelHtml5", "print"]
+    default_length_menu = [5, 10, 20]
+
+    layout = layout or default_layout
+    buttons = buttons or default_buttons
+    length_menu = length_menu or default_length_menu
+
+    opt.layout = layout
```
MrCreosote, line 40: This looks like a race condition theoretically

src/notebook_utils/notebook_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        length_menu: list = None
+) -> None:
+    """
+    Display a pandas DataFrame using itables.
+    iTables project page: https://github.com/mwouts/itables
+
+    :param df: a pandas DataFrame
+    :param layout: layout options, refer to https://datatables.net/reference/option/layout
+    :param buttons: buttons options, options refer to https://datatables.net/reference/button/
+    :param length_menu: length menu options
```
MrCreosote, line 19: I have no idea what length menu does based on reading the docs and the code

src/spark/utils.py
```diff
@@ -59,6 +63,10 @@ def _get_delta_lake_conf(
         "spark.hadoop.fs.s3a.path.style.access": "true",
         "spark.hadoop.fs.s3a.impl": "org.apache.hadoop.fs.s3a.S3AFileSystem",
         "spark.sql.catalogImplementation": "hive",
+        # SparkMonitor extension configuration
+        # https://github.com/swan-cern/sparkmonitor?tab=readme-ov-file#setting-up-the-extension
+        "spark.extraListeners": "sparkmonitor.listener.JupyterSparkMonitorListener",
+        "spark.driver.extraClassPath": f"{site_packages_path}/sparkmonitor/listener_{SCALA_VER}.jar",
```
MrCreosote, line 69: This jar is built in? I don't see any gradle changes

src/notebook_utils/notebook_utils.py
Tianhao-Gu: I added a TODO item. Since spark DF is not loaded in driver's memory before you call something like `show()`. I don't think any of the extension work with spark df. Unless we want to automatically convert spark df to pandas which might overload memory.

src/notebook_utils/notebook_utils.py
Tianhao-Gu: 👍 added a lock to prevent race condition.

src/notebook_utils/notebook_utils.py
Tianhao-Gu: 👍 added a reference.

src/spark/utils.py
Tianhao-Gu: it comes with pip sparkmonitor.

src/notebook_utils/notebook_utils.py
MrCreosote: Are we using the spark pandas api or regular pandas here?

src/notebook_utils/notebook_utils.py
MrCreosote: Ok, but is it ok to do that? I don't know if there are any repercussions to running what looks like a one time function many times

src/spark/utils.py
MrCreosote: :+1:

src/notebook_utils/notebook_utils.py
Tianhao-Gu: I think it's probably ok to do that. Seems like the func is setting pandas df styler and _repr_html_. If we just do it once at the beginning, it can be overwritten I think. Then I don't know the behavior of the below code.  
https://github.com/mwouts/itables/blob/main/src/itables/javascript.py#L75

src/notebook_utils/notebook_utils.py
Tianhao-Gu: > spark pandas api?

Do you mean something like spark_df.toPandas()? `spark_df.toPandas()` returns a regular pandas df.

src/notebook_utils/notebook_utils.py
MrCreosote: :+1:

src/notebook_utils/notebook_utils.py
MrCreosote: https://towardsdatascience.com/run-pandas-as-fast-as-spark-f5eefe780c45

Tianhao-Gu: it doesn't work for `df.to_pandas_on_spark()`

src/notebook_utils/notebook_utils.py
Tianhao-Gu: a `lock = RLock()` outside of `display_df`?

MrCreosote: yeah, just at the package level should work

src/notebook_utils/notebook_utils.py
MrCreosote: Ok, so this function should only be used in the case where the data is in a pandas DF, which means it's all in memory and localized to a single machine, e.g. relatively small?

src/notebook_utils/notebook_utils.py
Tianhao-Gu: 👍

MrCreosote: :+1:

src/notebook_utils/notebook_utils.py
Tianhao-Gu: yea that's correct. Maybe I should convert first few thousands of records of spark DF if a spark DF is given?

MrCreosote: I think at the minimum there should be a warning in the function docs about converting spark dfs to pandas dfs, as I assume doing that naively could pull a huge amount of data into memory on a single worker.

If you want to make any enhancements beyond that that's up to you but I'd do them in a different PR. A helper to `head` a spark df into an itable without pulling the rest of the data, where the maximum # of rows is 1K or something seems useful

MrCreosote: maybe with an offset arg as well

Tianhao-Gu: 👍 added a notice in the doc. I will look into how to properly read a spark df in later PRs.

MrCreosote: :+1:

MrCreosote (approved):

PR #32, opened 2024-06-11
add capability for users to include custom Spark config

Tianhao-Gu:


MrCreosote:
This seems dangerous - why is this necessary? Seems like a user could easily break the driver, set limits too high, reserve the entire cluster, etc

Tianhao-Gu:
> This seems dangerous - why is this necessary? Seems like a user could easily break the driver, set limits too high, reserve the entire cluster, etc

hmmm yea that's a concern. I was thinking this could give us more flexibilities to config spark session. Closing PR. We could park this feature if we need it.

PR #33, opened 2024-06-17
add user guide for accessing spark Jupyter lab

Tianhao-Gu:


docs/user_guide.md
```diff
@@ -0,0 +1,143 @@
...
+### 6. Closing the Spark Session:
+Please remember to close the Spark session when you are done with your work. This will release the resources and 
+prevent any memory leaks.
+
+```python
+spark.stop()
+```
+
+Please be aware that your session will automatically close after `1 hour`. Should you require an extension, simply invoke 
+`get_spark_session()` to initiate a new session.
```
MrCreosote, line 97: this means the prior session is still sitting around wasting resources, right?

docs/user_guide.md
```diff
@@ -0,0 +1,143 @@
...
+    monitor.send(msg)
+  File "/opt/bitnami/python/lib/python3.11/site-packages/sparkmonitor/kernelextension.py", line 57, in send
+    self.comm.send(msg)
+    ^^^^^^^^^
+AttributeError: 'ScalaMonitor' object has no attribute 'comm'
+```
+
+This error occurs when there is an issue with the SparkMonitor kernel extension, specifically related to the 
+communication between the kernel and the front end. The error indicates that the comm attribute is missing from the 
+ScalaMonitor object.
```
MrCreosote, line 133: What causes this?

Tianhao-Gu: From the extension developer:

`This is a difficult error to debug; the kernel extension is complaining that the Scala Listener has no comms; this is usually because the listener wasn't able to find a Spark Instance.`
https://github.com/itsjafer/jupyterlab-sparkmonitor/issues/16

docs/user_guide.md
Tianhao-Gu: no. after 1 hr, the prior session will be closed by the timer.

MrCreosote: But for that hour it is, that's what I'm talking about

docs/user_guide.md
MrCreosote: Hmm, that ticket says the bug is fixed

docs/user_guide.md
Tianhao-Gu: yea. for the hour the session is open, it's taking resources.

MrCreosote: I suppose we could keep track of open sessions and close them when a new session is requested but that's probably overkill for now

docs/user_guide.md
Tianhao-Gu: yea. It didn't happen very often. And restart kernel always works.

MrCreosote (approved):

PR #35, opened 2024-06-24
add jupyterlab git extension

Tianhao-Gu:


docker-compose.yaml
```diff
@@ -163,7 +163,7 @@ services:
   postgres:
     image: postgres:16.3
     restart: always
-    container_name: postgres
+    container_name: postgres_spark
```
MrCreosote, line 166: I don't see any reason to do this - seems fine as is.

Tianhao-Gu: so that we don't need to manually delete `postgres` container if we run docker-compose on this repo and on hive repo since they both have a `postgres` container.

Tianhao-Gu: actually, I revert the container name changes. We need to update `POSTGRES_URL` for every other dependencies. It's not worth it.

MrCreosote: Plus I suspect we'll need to use the hive only compose file rarely at best.

Pipfile
```diff
@@ -15,6 +15,7 @@ sparkmonitor = "==3.0.2"
 itables = "==2.1.1"
 PyMySQL = "==1.1.1"
 urllib3 = "==2.2.2"
+jupyterlab-git = "==0.50.1"
```
MrCreosote, line 18: Do docs need to be added to explain to users how to make use of this?

Tianhao-Gu: I think it's pretty self-explanatory as the extension added an UI component. Users don't need to do any extra steps to enable this. 

Screenshot 2024-06-24 at 11 40 56 AM

docker-compose.yaml
MrCreosote: if we're going to change it we should change it there

Pipfile
MrCreosote: :+1:

MrCreosote (approved):

PR #47, opened 2024-07-12
Allow for caching chown steps

MrCreosote:
These changes move chown steps earlier in the build so they can be cached. In some (my) cases, the current one liner chown can take upwards of 8 minutes x 5 containers when running docker compose up --build. Splitting them up allows for rarely changed chown targets to be cached and reused.


Tianhao-Gu (approved):
👍

PR #49, opened 2024-07-22
enable collaboration extension

Tianhao-Gu:
Screenshot 2024-07-21 at 11 48 34 AM

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,6 +16,7 @@ itables = "==2.1.4"
 PyMySQL = "==1.1.1"
 urllib3 = "==2.2.2"
 jupyterlab-git = "==0.50.1"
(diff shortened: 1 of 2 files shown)


MrCreosote:
Well that was easy.

Are any instructions needed for users?

How does it work given multiple people are making connections to the same port?

Tianhao-Gu:
> Well that was easy.
> 
> Are any instructions needed for users?
> 
> How does it work given multiple people are making connections to the same port?

yea. The lib makes it pretty easy to use. There is really no instructions needed I feel. It works exactly like google docs, you will see someone else is working on the notebook and any changes made by other person is instantly showing to your notebook.

MrCreosote (approved):

PR #60, opened 2024-08-05
add port 9003 to the document

Tianhao-Gu:


docs/minio_guide.md
```diff
@@ -10,6 +10,7 @@ interacting with MinIO.
 
     ```bash
     ssh -f -N -L 9002:ci07:9002 <ac.anl_username>@login1.berkeley.kbase.us 
+    ssh -f -N -L 9003:ci07:9003 <ac.anl_username>@login1.berkeley.kbase.us
```
MrCreosote, line 13: If the plan is to give out access keys that can't log into the UI, is this still necessary?

Tianhao-Gu: yea. I guess just in case they need to. It won't hurt I guess.

MrCreosote: Ok, maybe make a note that the UI isn't available to ordinary access creds so people aren't confused when they can't log in?

Tianhao-Gu: 👍

MrCreosote (approved):

PR #61, opened 2024-08-05
dependabot - bump lib version

Tianhao-Gu:


Pipfile
```diff
@@ -4,24 +4,24 @@ verify_ssl = true
...
-jupyterlab= "==4.2.3"
+jupyterlab= "==4.2.4"
 pyspark= "==3.5.1"
-boto3 = "==1.34.142"
+boto3 = "==1.34.152"
 minio = "==7.2.7"
 delta-spark = "==3.2.0"  # should match JAR version (DELTA_SPARK_VER) specified in the Dockerfile
 pandas = "==2.2.2"
-pyarrow = "==16.1.0"
+pyarrow = "==17.0.0"
```
MrCreosote, line 13: What is this for again?

Tianhao-Gu: This is similar to Pandas, but it's an Apache version. I cannot recall why I added this lib. We could try to remove it and if anything break we added back. But it really don't hurt to keep it there.

MrCreosote (approved):

PR #64, opened 2024-08-06
add minio read/write user and policy

Tianhao-Gu:
Tested creating delta tables with r/w user locally with dev jupyterlab. The read/write user can also access the UI but will only be able to see the buckets assigned to them. I will also hide minio admin user from dev juptyerlab in Rancher. 

Screenshot 2024-08-06 at 1 23 40 PM


scripts/minio_create_bucket_entrypoint.sh
```diff
@@ -18,15 +18,22 @@ fi
...
 # make yarn user
 mc admin user add minio yarnuser yarnpass
 mc admin policy attach minio yarn-write-policy --user=yarnuser
 echo 'YARN user and policy set'
+
+# make read/write user
+mc admin user add minio minio-readwrite minio123
+mc admin policy attach minio cdm-read-write-policy --user=minio-readwrite
+mc admin policy attach minio yarn-write-policy --user=minio-readwrite
+echo 'CDM read-write user and policy set'
```
MrCreosote, line 39: So the idea is that the CI dev notebook will also only get read / write creds?

MrCreosote (approved):

scripts/minio_create_bucket_entrypoint.sh
Tianhao-Gu: yea. So that we can hide minio admin creds.

PR #65, opened 2024-08-07
plotly extension + user lib installation bug

Tianhao-Gu:
Enable Plotly plot + Users can now install any package they need using pip.

Screenshot 2024-08-06 at 3 58 04 PM


Dockerfile
```diff
@@ -29,19 +29,22 @@ ENV GRADLE_JARS_DIR=gradle_jars
...
 # install pipenv
 RUN pip3 install pipenv
 
 # install python dependencies
 COPY Pipfile* ./
 RUN pipenv sync --system
 
+# Update permissions on the Bitnami directory after Pipenv installs
+# to ensure the Spark user can install packages properly within notebook.
+RUN chown -R spark_user:spark /opt/bitnami
```
MrCreosote, line 41: In my hands the bitnami chown takes a really long time and now will rerun anytime we change the python packages. Why does it need to be moved forward in the build? pip shouldn't touch it, right? 

Same question for the yarn setup below, although that's fast

Tianhao-Gu: I think pipenv run by `root` changes owners of some of python lib dir. As a result, users sometimes failed use pip install.

MrCreosote: The python lib dir is in /opt/bitnami?

MrCreosote: What if we ran pipenv with the spark user?

Tianhao-Gu: Cannot run pipenv with spark user. python lib is in /opt/bitnami. Maybe docker works even differently on linux and mac? `chown` is cached with mac and it only takes less than a minutes to build for me.

MrCreosote: > Cannot run pipenv with spark user. 

Why not?

> chown is cached with mac and it only takes less than a minutes to build for me.

It can't be if the pip install changes given it's after that step. It takes 5-10 minutes for me

MrCreosote: Can you use `--extra-pip-args --user`?

Tianhao-Gu: > > Cannot run pipenv with spark user.
> 
> Why not?
> Permission issue
> > chown is cached with mac and it only takes less than a minutes to build for me.
> 
> It can't be if the pip install changes given it's after that step. It takes 5-10 minutes for me
It doesn't happen to me. But I am ok with that given users cannot run pip install with current setup.

Tianhao-Gu: > Can you use `--extra-pip-args="--user"`?

I don't believe it's worth the effort. Why risk potential errors for users just to expedite local development?

MrCreosote: > Permission issue

See post above

> It doesn't happen to me. But I am ok with that given users cannot run pip install with current setup.

I mean of course you're ok with it, it's fast for you.

MrCreosote: So hold on - if pipenv is installing into /opt/bitnami, and /opt/bitnami is owned by the spark user, how can there be a permissions issue if pipenv is running as the spark user?

Tianhao-Gu: > So hold on - if pipenv is installing into /opt/bitnami, and /opt/bitnami is owned by the spark user, how can there be a permissions issue if pipenv is running as the spark user?

It was run by root with `pipenv sync `

MrCreosote: You said there were permissions issues when running it with the spark user though?

Tianhao-Gu: > You said there were permissions issues when running it with the spark user though?

You cannot run pipenv with spark user. 
```
 => ERROR [spark-master  8/24] RUN pip3 install pipenv                     1.0s 
------                                                                          
 > [spark-master  8/24] RUN pip3 install pipenv:                                
0.232 WARNING: The directory '/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
0.232 Defaulting to user installation because normal site-packages is not writeable
0.249 DEPRECATION: Loading egg at /opt/bitnami/python/lib/python3.11/site-packages/pip-23.3.2-py3.11.egg is deprecated. pip 24.3 will enforce this behaviour change. A possible replacement is to use pip for package installation. Discussion can be found at https://github.com/pypa/pip/issues/12330
0.348 Collecting pipenv
0.467   Downloading pipenv-2024.0.1-py3-none-any.whl.metadata (19 kB)
0.495 Collecting certifi (from pipenv)
0.523   Downloading certifi-2024.7.4-py3-none-any.whl.metadata (2.2 kB)
0.525 Requirement already satisfied: setuptools>=67 in /opt/bitnami/python/lib/python3.11/site-packages (from pipenv) (68.2.2)
0.526 Requirement already satisfied: virtualenv>=20.24.2 in /opt/bitnami/python/lib/python3.11/site-packages (from pipenv) (20.26.3)
0.532 Requirement already satisfied: distlib<1,>=0.3.7 in /opt/bitnami/python/lib/python3.11/site-packages (from virtualenv>=20.24.2->pipenv) (0.3.8)
0.532 Requirement already satisfied: filelock<4,>=3.12.2 in /opt/bitnami/python/lib/python3.11/site-packages (from virtualenv>=20.24.2->pipenv) (3.15.3)
0.533 Requirement already satisfied: platformdirs<5,>=3.9.1 in /opt/bitnami/python/lib/python3.11/site-packages (from virtualenv>=20.24.2->pipenv) (4.2.2)
0.567 Downloading pipenv-2024.0.1-py3-none-any.whl (3.2 MB)
0.746    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 3.2/3.2 MB 18.3 MB/s eta 0:00:00
0.786 Downloading certifi-2024.7.4-py3-none-any.whl (162 kB)
0.791    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 163.0/163.0 kB 121.1 MB/s eta 0:00:00
0.825 Installing collected packages: certifi, pipenv
0.826 ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/.local/lib'
0.826 Check the permissions.
0.826 
------
```

Tianhao-Gu: I did chown bitnami dir before pipenv 

```
RUN chown -R spark_user:spark /opt/bitnami

USER spark_user

# install pipenv
RUN pip3 install pipenv

# install python dependencies
COPY Pipfile* ./
RUN pipenv sync --system
```

MrCreosote: Oh, I bet because `$HOME` is set to `/`. I'd guess setting $HOME to `/home/spark` would work

Tianhao-Gu: > Oh, I bet because `$HOME` is set to `/`. I'd guess setting $HOME to `/home/spark` would work
> 
> edit: you can add the `-d` flag to `useradd` to make it create the home dir appropriately

we do all those tricks risking errors on the users' end just to get image build a few minutes faster on a linux machine?

MrCreosote: Setting up a proper home directory for a user isn't a trick, it's standard practice, and if I'm right, the lack of a proper homedir is what's causing this error (and will probably cause other errors as well, since the spark user is trying to use the root dir as their home dir where they don't have write)

In a noninteractive container you probably don't need a home dir, but in an interactive container like this I would expect that having a non-writeable home dir would cause all sorts of problems

Tianhao-Gu: > Setting up a proper home directory for a user isn't a trick, it's standard practice, and if I'm right, the lack of a proper homedir is what's causing this error (and will probably cause other errors as well, since the spark user is trying to use the root dir as their home dir where they don't have write)
> 
> In a noninteractive container you probably don't need a home dir, but in an interactive container like this I would expect that having a non-writeable home dir would cause all sorts of problems

You don't think `RUN chown -R spark_user:spark /opt/bitnami` is the safest way to get rid of permission issues for users? Tricks/practice you can name it. But we do extra works in exchange of a less safer way to fix user's problem.

MrCreosote: OOB discussion is that @Tianhao-Gu will give the writable spark home dir approach a try

Tianhao-Gu: @MrCreosote Setting HOME to bitnami dir. That seems to work.

MrCreosote: Hmm, I'd be worried about writes to the home directory potentially clobbering bitnami files. I'd recommend a separate home dir

Dockerfile
```diff
@@ -73,4 +66,14 @@ RUN chown -R spark_user:spark $CDM_SHARED_DIR
...
+# Set home directory for the spark user so that pipenv can install packages properly with spark user permissions
+ENV HOME=/opt/bitnami
+
+# install pipenv
+RUN pip3 install pipenv
+
+# install python dependencies
+COPY Pipfile* ./
+RUN pipenv sync --system
+
```
MrCreosote, line 78: With the pip installs this late the in file, they'll get reinstalled any time `src` changes.

Dockerfile
Tianhao-Gu: using other HOME will result permission error when pip install in the notebook. 
```
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/opt/bitnami/python/share/man/man1/ttx.1'
Consider using the `--user` option or check the permissions.
```

MrCreosote: How is that possible when that directory is chowned to spark?

Tianhao-Gu: no clue. That's why I think the safest way is to run `RUN chown -R spark_user:spark /opt/bitnami` after we done modifying that dir.

MrCreosote: This makes no sense whatsoever. Why would setting $HOME to /opt/bitnami fix a permissions error?

MrCreosote: Ok, how about we leave home as /opt/bitnami for now but make a ticket to look into this weirdness at some point. At least we'll have a record if something similar starts happening

Dockerfile

Dockerfile
Tianhao-Gu: 👍  https://github.com/kbase/cdm-jupyterhub/issues/66

MrCreosote (approved):

PR #74, opened 2024-08-27
add developer guide for loading data

Tianhao-Gu:


docs/dev_guide.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Accessing JupyterLab Development Environment
+
+### 1. Create SSH Tunnel:
+
+Execute the following command to create an SSH tunnel to the remote server (`login1.berkeley.kbase.us`):
+
+```bash
+ssh -f -N -L localhost:44041:10.58.2.201:4041 <ac.anl_username>@login1.berkeley.kbase.us
+```
```
MrCreosote, line 11: I thought we already had docs like this somewhere?

docs/dev_guide.md
```diff
@@ -0,0 +1,60 @@
...
+## Accessing MinIO
+Please refer to the [MinIO Guide](minio_guide.md) for instructions on accessing MinIO.
+
+### Read/Write MinIO username and password
+Get the MinIO username and password with read/write permission from the above development JupiterLab environment.
+```python
+import os
+minio_username, minio_password = os.environ['MINIO_ACCESS_KEY'], os.environ['MINIO_SECRET_KEY']
+print(f"MinIO username: {minio_username},\nMinIO password: {minio_password}")
+```
```
MrCreosote, line 32: These are r/w not admin right?

docs/dev_guide.md
```diff
@@ -0,0 +1,60 @@
...
+Delta table files are Parquet files generated by Spark during the creation of a table.
+* Bucket name: `namespace_name`-delta
+
+## Loading Notebooks
+Please create a corresponding loading notebook for each namespace in the `data-loading-notebooks` directory.
+
+Please use the existing loading notebooks as examples.
+
+🚨 **Please DO NOT rerun the loading notebooks in the development environment. Instead, create a new notebook for each 
+new namespace and manually verify the data loading process.**
```
MrCreosote, line 53: I'm not totally sure what this means. If you need to reload the data shouldn't you reuse the notebook?

docs/dev_guide.md
Tianhao-Gu: yea. Similar instructions in  the user's guide. https://github.com/kbase/cdm-jupyterhub/blob/main/docs/user_guide.md
But dev has different port number. That's also the reason I didn't make it as detailed as the user's guide.

docs/dev_guide.md
Tianhao-Gu: Updated the note to make it more clear. I was trying to prevent people running existing loading notebook and accidentally update existing tables.

docs/dev_guide.md
MrCreosote: Maybe link to the user guide for more info?

docs/dev_guide.md
MrCreosote: :+1:

docs/dev_guide.md
MrCreosote: Is there anything that should be added about ensuring the notebook works when overwriting as opposed to loading clean? I'm not sure there's a guarantee that a notebook can be run without dropping the tables first

docs/dev_guide.md
Tianhao-Gu: 👍

docs/dev_guide.md
Tianhao-Gu: I think that's fine for now. I haven't thoroughly tested the process of reloading or altering tables yet. For now, I just want to ensure people only load/update their own tables and avoid modifying others' tables. I've observed that people often rerun notebooks to check the output for convenience.

MrCreosote (approved):

PR #75, opened 2024-08-29
using ghcr.io/kbase img for standalone spark deployment

Tianhao-Gu:
Screenshot 2024-08-29 at 10 15 27 AM

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -7,6 +7,8 @@ services:
   yarn-resourcemanager:
     image: ghcr.io/kbase/cdm-prototype-yarn:pr-8
     container_name: yarn-resourcemanager
+    # Images from the ghcr.io/kbase registry are exclusively available for Linux/AMD64 platforms.
+    platform: linux/amd64
     ports:
       - 8088:8088  # web ui
     environment:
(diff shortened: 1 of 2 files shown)


MrCreosote:
This is just the prelude to pulling things out of the image to slim it down? What can you actually remove? You still need the spark libs right?

docker-compose.yaml
```diff
@@ -7,6 +7,8 @@ services:
   yarn-resourcemanager:
     image: ghcr.io/kbase/cdm-prototype-yarn:pr-8
     container_name: yarn-resourcemanager
+    # Images from the ghcr.io/kbase registry are exclusively available for Linux/AMD64 platforms.
+    platform: linux/amd64
```
MrCreosote, line 11: Does this fix the mac issues?

MrCreosote (approved):
LGTM in any case

docker-compose.yaml
Tianhao-Gu: No. Without this, docker compose won't run on Mac machines since it cannot find the arm64 version. I used to manually modify docker-compose file before running it.

Tianhao-Gu:
> This is just the prelude to pulling things out of the image to slim it down? What can you actually remove? You still need the spark libs right?

I will test and see if I can remove any spark libs in the next PRs. I want to do this little by little so that I can test in CI since local yarn doesn't work for me.

PR #76, opened 2024-09-03
remove sparkmonitor

Tianhao-Gu:
Remove SparkMonitor based on feedback from the demo.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -5,13 +5,14 @@ name = "pypi"
 
 [packages]
 jupyterlab= "==4.2.4"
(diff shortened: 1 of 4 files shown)


MrCreosote (approved):
Do you want to make this a toggle vs a straight removal? In the future users might be able to choose whether it's enabled or not if so

Tianhao-Gu:
> Do you want to make this a toggle vs a straight removal? In the future users might be able to choose whether it's enabled or not if so

I think it's fine to remove it for now. If needed, we can easily reintroduce it later.

PR #77, opened 2024-09-03
add JUPYTER_MODE for JupyterHub

Tianhao-Gu:


docker-compose.yaml
```diff
@@ -179,6 +181,65 @@ services:
     volumes:
       - ./cdr/cdm/jupyter/user_shared_workspace:/cdm_shared_workspace/user_shared_workspace
 
+  dev_jupyterhub:
```
Tianhao-Gu, line 184: We probably only need one jupyterhub instance eventually and control permission based on different user groups. But I haven't completely sort it out yet.

Dockerfile
```diff
@@ -42,13 +42,25 @@ RUN pipenv sync --system
...
 RUN chown -R spark_user:spark /.jupyter
 
+# Set up Jupyter Hub directories
+ENV JUPYTERHUB_CONFIG_DIR=/srv/jupyterhub
+RUN mkdir -p ${JUPYTERHUB_CONFIG_DIR}
+COPY ./src/notebook_utils/startup.py ${JUPYTERHUB_CONFIG_DIR}/startup.py
+RUN chown -R spark_user:spark ${JUPYTERHUB_CONFIG_DIR}
+
+# Jupyter Hub user home directory
+RUN mkdir -p /home
```
MrCreosote, line 59: This is the standard home root dir for all users. I'd pick something else just in case

docker-compose.yaml
```diff
@@ -179,6 +181,65 @@ services:
...
+        - S3_YARN_BUCKET=yarn
+        - MAX_EXECUTORS=4
+        - POSTGRES_USER=hive
+        - POSTGRES_PASSWORD=hivepassword
+        - POSTGRES_DB=hive
+        - POSTGRES_URL=postgres:5432
+        - USAGE_MODE=dev
+      volumes:
+        - ./cdr/cdm/jupyter:/cdm_shared_workspace
+        - ./cdr/cdm/jupyter/home:/home
```
MrCreosote, line 212: I'd make this specific to jhub to avoid confusion

docker-compose.yaml
```diff
@@ -179,6 +181,65 @@ services:
...
+      - MINIO_URL=http://minio:9002
+      - MINIO_ACCESS_KEY=minio-readonly
+      - MINIO_SECRET_KEY=minio123
+      - S3_YARN_BUCKET=yarn
+      - JUPYTER_MODE=jupyterhub
+      - MAX_EXECUTORS=4
+      - POSTGRES_USER=hive
+      - POSTGRES_PASSWORD=hivepassword
+      - POSTGRES_DB=hive
+      - POSTGRES_URL=postgres:5432
```
MrCreosote, line 239: Not really the point of this PR but it makes me nervous that users have access to these creds...

docker-compose.yaml
```diff
@@ -179,6 +181,65 @@ services:
...
+    volumes:
+      - ./cdr/cdm/jupyter/home:/home
```
MrCreosote, line 241: No shared workspace?

MrCreosote (commented):
I take it this is just the initial hub setup and there's more work to come

Dockerfile

docker-compose.yaml

docker-compose.yaml
Tianhao-Gu: no. Each user will have their own dir and python environment.

docker-compose.yaml
MrCreosote: Even that would make me nervous. It means that any user could blow away the hive tables

docker-compose.yaml
MrCreosote: Should `dev_jupyterhub` have the shared workspace mounted in that case?

Dockerfile

docker-compose.yaml

docker-compose.yaml
Tianhao-Gu: both instances mount `- ./cdr/cdm/jupyter/jupyterhub/users_home:/jupyterhub/users_home`. Each user will have their own dir inside of `/jupyterhub/users_home` and create notebook there.

MrCreosote: My point is that `dev-jupyterhub` *also* mounts `/cdr/cdm/jupyter:/cdm_shared_workspace`. I'm asking if that line needs to be there or if it should be removed to match `user_juypterhub`

Tianhao-Gu: It should be there, at least for now. There are postgres files there. Dev environment should have access to them.

docker-compose.yaml
Tianhao-Gu: I updated the todo item to 
```
# TODO: create postgres user r/ only read access to the hive tables
```

docker-compose.yaml
MrCreosote: I thought we were running a separate postgres server? What are the local files?

docker-compose.yaml
MrCreosote: I guess another question is if users and devs are all in the same jupyterhub instance is it possible to have different environments

docker-compose.yaml
Tianhao-Gu: there is a `hive_metastore` and `cdm-postgres` dir.

docker-compose.yaml
Tianhao-Gu: Users won't since they don't have minIO write permission. I am hoping it can be configured. But TBD.

docker-compose.yaml
Tianhao-Gu: I don't see any issues with using `- ./cdr/cdm/jupyter:/cdm_shared_workspace` for the dev environment.

docker-compose.yaml
MrCreosote: Ok, so users = read only, devs = write, essentially.

docker-compose.yaml
MrCreosote: I'm not saying there's an issue, I just wasn't sure why it was there. can you remind me what's in the hive_metastore and cdm-postgres dirs? Are they both still relevant with the standalone postgres service?

docker-compose.yaml
MrCreosote: If we could ever get the remote metastore working that might provide some protection as well, not sure

docker-compose.yaml
Tianhao-Gu: I think cdm-postgres is not longer used, it's not getting updated for 2 months. 
In `hive_metastore` there are <db_name>.db files. I think they are used when users run spark select statement. 

Screenshot 2024-09-03 at 3 01 27 PM

MrCreosote: Ok, maybe cdm-postgres was the local DB, and hive_metastore is cache data?

MrCreosote (approved):

docker-compose.yaml
Tianhao-Gu: not exactly sure. But I want to worry about those later.

PR #78, opened 2024-09-05
setup jupyterhub config

Tianhao-Gu:


scripts/spawn_notebook.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+USERNAME=${JUPYTERHUB_USER}
+
+echo "Starting Jupyter Notebook for user: $USERNAME"
+cd $JUPYTERHUB_USER_HOME/$USERNAME
```
Tianhao-Gu, line 6: User home dir will also be created in `custom_spawner.py` in future PRs.

Dockerfile
```diff
@@ -82,6 +98,10 @@ ENV CDM_SHARED_DIR=/cdm_shared_workspace
 RUN mkdir -p ${CDM_SHARED_DIR} && chmod -R 777 ${CDM_SHARED_DIR}
 RUN chown -R spark_user:spark $CDM_SHARED_DIR
 
+# Allow spark_user to use sudo without a password
```
Tianhao-Gu, line 101: Configuring JupyterHub with a non-root user is a lot challenging, as this user must manage other system users (home dir, virtual env, etc). JupyterHub requires the presence of system users for its operations.

config/jupyterhub_config.py
```diff
@@ -0,0 +1,26 @@
...
+import os
+
+from jupyterhub_config.custom_spawner import VirtualEnvSpawner
+
+c = get_config()
+
+# Set the authenticator class
+# TODO: Change the authenticator class to a secure one (e.g. GitHubOAuthenticator)
+c.JupyterHub.authenticator_class = 'jupyterhub.auth.DummyAuthenticator'
+c.Authenticator.allowed_users = {'spark_user', 'test_user1', 'test_user2'}
```
Tianhao-Gu, line 15: Non-existent system users (`test_user1`, `test_user2`) will be automatically created later in `custom_spawner.py`.

Tianhao-Gu: Eventually I want to switch to OAuth and phase out user/password auth completely. Maybe only allow admin users to use password.

MrCreosote: Presumably long term we'd swap out all the auth systems for KBase auth

Tianhao-Gu: yea. That's the plan to use KBase auth eventually.

Dockerfile
```diff
@@ -82,6 +98,10 @@ ENV CDM_SHARED_DIR=/cdm_shared_workspace
...
+# TODO: use `sudospawner` in JupyterHub to avoid this (https://jupyterhub.readthedocs.io/en/stable/howto/configuration/config-sudo.html)
+RUN echo "spark_user ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
```
MrCreosote, line 103: How do you ensure users can't access this account?

config/jupyterhub_config.py
```diff
@@ -0,0 +1,26 @@
+"""
+This is the JupyterHub configuration file. It is used to configure the JupyterHub server.
+Refer to the JupyterHub documentation for more information:
+https://jupyterhub.readthedocs.io/en/latest/tutorial/getting-started/config-basics.html
+"""
+import os
+
+from jupyterhub_config.custom_spawner import VirtualEnvSpawner
+
+c = get_config()
```
MrCreosote, line 10: Where does the `get_config` name come from? I'd expect this to throw a NameError

config/jupyterhub_config.py
```diff
@@ -0,0 +1,26 @@
...
+
+# Set the authenticator class
+# TODO: Change the authenticator class to a secure one (e.g. GitHubOAuthenticator)
+c.JupyterHub.authenticator_class = 'jupyterhub.auth.DummyAuthenticator'
+c.Authenticator.allowed_users = {'spark_user', 'test_user1', 'test_user2'}
+c.DummyAuthenticator.password = os.environ['JUPYTERHUB_ADMIN_PASSWORD']
```
MrCreosote, line 16: is there an authenticator that allows setting a password per user?

scripts/spawn_notebook.sh
```diff
@@ -0,0 +1,9 @@
...
+
+# Start the notebook server with current user
+exec jupyterhub-singleuser "$@"
```
MrCreosote, line 9: This looks like a shell injection vulnerability

MrCreosote (commented):
I don't really understand enough about how jupyterhub works to properly review this. What prevents user A from reading from or writing to user B's environment?

scripts/spawn_notebook.sh
Tianhao-Gu: 👍

config/jupyterhub_config.py
Tianhao-Gu: I think there is but I want to explore that in the future.

config/jupyterhub_config.py
Tianhao-Gu: It's imported automatically by jupyterhub.

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
MrCreosote: E.g. it monkey patches the config file?

scripts/spawn_notebook.sh
MrCreosote: :+1:

Dockerfile
Tianhao-Gu: users logged in as individual system user. I don't know they can have access to spark_user.

Dockerfile
```diff
@@ -42,22 +54,26 @@ RUN pipenv sync --system
 
 RUN chown -R spark_user:spark /opt/bitnami
 
-# Set up Jupyter Lab directories
+# Set up JupyterLab directories
```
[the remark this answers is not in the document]
Tianhao-Gu: yea. I think so.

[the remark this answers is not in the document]
MrCreosote: :+1:

Dockerfile
MrCreosote: So you have to make a user account for each user?

Tianhao-Gu:
> I don't really understand enough about how jupyterhub works to properly review this. What prevents user A from reading from or writing to user B's environment?

Upon logging in, we need to ensure user's home directory is properly created and permissioned which will be done by `VirtualEnvSpawner` class.

Dockerfile
Tianhao-Gu: Yes. Jupyterhub operates only on system users.

MrCreosote: Ok, so what the spawner is going to do it make a new user if the user & set up home if it doesn't exist, and rely on the linux permissions to make sure they can't read other user's envs?

config/jupyterhub_config.py
Tianhao-Gu: hmmm not exactly sure. But it's ready to use by Jupyterhub automatically. 
https://jupyterhub.readthedocs.io/en/stable/reference/config-reference.html

Dockerfile
Tianhao-Gu: Yea. Will implement those in the future PRs.

config/jupyterhub_config.py
MrCreosote: can you link to that in the config as well as the link you already have

Dockerfile
MrCreosote: :+1:

config/jupyterhub_config.py
Tianhao-Gu: 👍

MrCreosote: :+1:

MrCreosote (approved):

PR #79, opened 2024-09-06
add create system user

Tianhao-Gu:


src/jupyterhub_config/custom_spawner.py
```diff
@@ -14,4 +16,43 @@ def start(self):
...
+            # Create the user
+            self.log.info(f'Creating system user: {username}')
+            useradd_cmd = ['sudo', 'useradd', '-r']
+
+            if group:
+                # Check if the group exists, create if necessary
+                group_check = subprocess.run(['getent', 'group', group], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                if group_check.returncode != 0:
+                    self.log.info(f'Group {group} does not exist, creating it.')
+                    subprocess.run(['sudo', 'groupadd', group], check=True)
```
MrCreosote, line 47: There's no other case where the returncode could be nonzero?

src/jupyterhub_config/custom_spawner.py
```diff
@@ -14,4 +16,43 @@ def start(self):
         variables, and sets the notebook directory before starting the server.
         """
 
-        return super().start()
\ No newline at end of file
+        username = self.user.name
```
MrCreosote, line 19: I assume this is coming from the superclass

src/jupyterhub_config/custom_spawner.py
```diff
@@ -14,4 +16,43 @@ def start(self):
...
+            result = subprocess.run(['id', username], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            if result.returncode == 0:
+                self.log.info(f'User {username} already exists')
+                return
+
+            # Create the user
+            self.log.info(f'Creating system user: {username}')
+            useradd_cmd = ['sudo', 'useradd', '-r']
+
+            if group:
```
MrCreosote, line 42: What is the plan around adding all the users to the jupyterhub group?

src/jupyterhub_config/custom_spawner.py
```diff
@@ -14,4 +16,43 @@ def start(self):
...
+                    self.log.info(f'Group {group} does not exist, creating it.')
+                    subprocess.run(['sudo', 'groupadd', group], check=True)
+                else:
+                    self.log.info(f'Group {group} already exists')
+
+                useradd_cmd.extend(['-g', group])
+
+            useradd_cmd.append(username)
+
+            subprocess.run(useradd_cmd, check=True)
```
MrCreosote, line 55: There's a potential race condition here if the same user tries to spawn a hub (?) repeatedly

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -1 +1,83 @@
...
+
+
+# Test when the user already exists
+@patch('subprocess.run')
+def test_ensure_system_user_already_exists(mock_run, mock_spawner):
+    # Mock 'id' command to simulate user already exists
+    mock_run.return_value.returncode = 0
+
+    username = 'testuser'
+    mock_spawner._ensure_system_user(username)
```
MrCreosote, line 24: I assume there's a reason you're calling this instead of start? Calling internal functions is generally not good practice

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -1 +1,83 @@
...
+
+import pytest
+
+from jupyterhub_config.custom_spawner import VirtualEnvSpawner
+
+
+@pytest.fixture
+def mock_spawner():
+    spawner = VirtualEnvSpawner()
+    spawner.log = MagicMock()  # Mock the logger
```
MrCreosote, line 13: Hmm, if the superclass logger is not at this location or has a different interface this won't catch it

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -1 +1,83 @@
...
+# Test when the group and user need to be created
+@patch('subprocess.run')
+def test_ensure_system_user_create_group_and_user(mock_run, mock_spawner):
+    # Mock the 'id' and 'getent' command returncodes
+    def mock_run_side_effect(cmd, *args, **kwargs):
+        if cmd == ['id', 'testuser']:
+            return MagicMock(returncode=1)  # User does not exist
+        elif cmd == ['getent', 'group', 'testgroup']:
+            return MagicMock(returncode=2)  # Group does not exist
+        return MagicMock(returncode=0)  # Successful creation
```
MrCreosote, line 39: Instead of this you could just set `side_effect=[1, 2, 0]` and then `assert_has_calls(<the arguments above in order>)`

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -1 +1,83 @@
...
+
+    mock_run.side_effect = mock_run_side_effect
+
+    username = 'testuser'
+    group = 'testgroup'
+    mock_spawner._ensure_system_user(username, group)
+
+    mock_spawner.log.info.assert_any_call(f'Creating system user: {username}')
+    mock_spawner.log.info.assert_any_call(f'Group {group} does not exist, creating it.')
```
MrCreosote, line 48: I would use `assert_has_calls` here

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -1 +1,83 @@
...
+    mock_run.side_effect = [
+        MagicMock(returncode=1),  # User does not exist
+        MagicMock(returncode=0)  # User created successfully
+    ]
+
+    username = 'testuser'
+    mock_spawner._ensure_system_user(username)
+
+    mock_spawner.log.info.assert_any_call(f'Creating system user: {username}')
+    mock_run.assert_called_with(['sudo', 'useradd', '-r', username], check=True)
```
MrCreosote, line 66: I would use assert_called_once_with for the 2 lines above

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -1 +1,83 @@
...
+    mock_run.side_effect = [
+        MagicMock(returncode=1),
+        CalledProcessError(1, 'useradd')
+    ]
+
+    with pytest.raises(ValueError, match="Failed to create system user"):
+        mock_spawner._ensure_system_user('testuser')
+
+    mock_run.assert_any_call(['id', 'testuser'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    mock_run.assert_any_call(['sudo', 'useradd', '-r', 'testuser'], check=True)
```
MrCreosote, line 83: assert_has_calls

src/jupyterhub_config/custom_spawner.py
MrCreosote: There's a race condition around group creation as well

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: https://docs.pytest.org/en/7.1.x/how-to/logging.html#caplog-fixture might be useful, although I don't know how the logger in the superclass works

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: There maybe. I don’t see I need to adjust creating group logic anyways.

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: Yea

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: I don’t have any. We are not there yet. Maybe assign group by OAuth or some type of user DB.

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: I don’t see any problem even if race condition happens.

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: So that we don’t need to mock start()

src/jupyterhub_config/custom_spawner.py
MrCreosote: in that case the group would exist, but then this code will force a failure by calling groupadd

src/jupyterhub_config/custom_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_spawner.py
MrCreosote: Ok, so why does the code add the users to the group then?

src/jupyterhub_config/custom_spawner.py
MrCreosote: The user wouldn't be able to get into the hub, depending on what instance they wind up pointed at

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: Why would you need to mock start?

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: But the tests don't check that it's working fine, that's my point. The logger could be at `logger` instead of `log` and the tests would pass

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: hmmm not something I am aware of. If group exist, `getent group` should return `0`.

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: think it's good to do so. Otherwise we end up with users with no group.

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: It calls `super().start()`. I don't think we need to call that for unit tests.

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: I don't get what you mean here. The code is using `self.log` from parent class.

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: This issue occurs only for new users logging in for the first time. And only if the user initiates multiple simultaneous logins, some logins succeed while others fail. We are concerning about that situation?

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: 👍

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: 👍

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: 👍

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: 👍

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: test also checks every `mock_spawner.log.info` calls.

src/jupyterhub_config/custom_spawner.py
MrCreosote: Not if there's some other failure that also returns a nonzero error code

src/jupyterhub_config/custom_spawner.py
MrCreosote: Ok, so this is an add a group because users need a group type thing. But that also means that users will be able to read and write each other's files by default unless the chmod 700 them. Is that the desired behavior?

src/jupyterhub_config/custom_spawner.py
MrCreosote: A first time user trying to log in and getting a failure doesn't seem like a good experience to me

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: Does it cause a problem to call it?

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: My point is that the test isn't actually testing the logger. The code under test is calling `super.log.info`, but that's getting mocked, and then the output of the mock is checked. That means if there's a bug in the code and the correct call was `super.logger.log_info`, for example, the test wouldn't catch it.

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: Looks good, but did you mean to remove the logger assertions?

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: Same as above, looks like the log assertions were removed completley

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: :+1:

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: :+1:

src/jupyterhub_config/custom_spawner.py
MrCreosote: You're checking to see if a group exists, and then if the code is nonzero, creating the group. That inherently assumes the only reason for a nonzero error code is that the group doesn't exist, that's the issue

Tianhao-Gu: what other reasons would you not create a group?

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: That's how NERSC does. KBase group members can read each other's files.

src/jupyterhub_config/custom_spawner.py
MrCreosote: If the group already exists but getent group returns a non-zero return code

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: `super().start()` does a lot of things. `_ensure_system_user` is what I want to test.

src/jupyterhub_config/custom_spawner.py
MrCreosote: Ok, if that's intentional :+1: 

I presume we'd have to change this if non-kbase users started using this system, but that's a ways off

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: Ok :

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: added back.

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: added back.

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: added lock but I think it's completely unnecessary.

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: I just really don't know how will that happen. Can you suggest how would you do it?

```
       One of the following exit values can be returned by getent:

       0      Command completed successfully.

       1      Missing arguments, or database unknown.

       2      One or more supplied key could not be found in the
              database.

       3      Enumeration not supported on this database.
```

src/jupyterhub_config/custom_spawner.py
MrCreosote: There's no possibility of this code being run concurrently?

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: :+1:

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: :+1:

src/jupyterhub_config/custom_spawner.py
MrCreosote: :+1:

Tianhao-Gu: It's possible. The  file lock will still work right?

src/jupyterhub_config/custom_spawner.py
MrCreosote: Hmm, highlighted lines are wrong, the first comment should be referring to the getent command

src/jupyterhub_config/custom_spawner.py
MrCreosote: yeah, I made that now deleted comment before I saw it was a file lock rather than a thread lock

src/jupyterhub_config/custom_spawner.py
MrCreosote: Nope, seems like the answer to the OP is no, but the OP was pointing at the wrong line of code which maybe caused some of this confusion

MrCreosote: :+1:

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: hmmm under what situation will that happen? I am more confused about `super.logger.log_info`. There is no `super.logger` involved.

Tianhao-Gu: you mean something goes wrong with `self.log.info` function call?

Tianhao-Gu: 👍 using caplog to capture the logger output.

MrCreosote: :+1:

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -1 +1,103 @@
...
+
+import pytest
+
+from jupyterhub_config.custom_spawner import VirtualEnvSpawner
+
+
+@pytest.fixture
+def mock_spawner():
+    spawner = VirtualEnvSpawner()
+    return spawner
```
MrCreosote, line 15: Seems like you could just get rid of this fixture and create it in the test methods, since there's no mocks any more

MrCreosote (approved):

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: oh yea. removed spawner fixture.

PR #81, opened 2024-09-09
create user env

Tianhao-Gu:


src/jupyterhub_config/custom_spawner.py
```diff
@@ -126,4 +133,38 @@ def _ensure_user_jupyter_directory(self, user_dir: Path):
...
+                raise
+        else:
+            self.log.info(f'Reusing virtual environment for {self.user.name}')
+
+    def _configure_environment(self, user_dir: Path, user_env_dir: Path, username: str):
+        """
+        Configure the environment variables for the user's session, including
+        the PATH and PYTHONPATH to use the virtual environment.
+        """
+        self.environment.update({key: value for key, value in os.environ.items() if key not in self.environment})
```
MrCreosote, line 161: This is the jupyterhub environment I guess? Is it not already populated with the os environment?

src/jupyterhub_config/custom_spawner.py
```diff
@@ -126,4 +133,38 @@ def _ensure_user_jupyter_directory(self, user_dir: Path):
...
+
+        self.environment['HOME'] = str(user_dir)
+        self.environment['PATH'] = f"{user_env_dir}/bin:{os.environ['PATH']}"
+        self.environment['PYTHONPATH'] = f"{user_env_dir}/lib/python3.11/site-packages:{os.environ['PYTHONPATH']}"
```
MrCreosote, line 165: Is PYTHONPATH guaranteed to be in the os env?

src/jupyterhub_config/custom_spawner.py
```diff
@@ -126,4 +133,38 @@ def _ensure_user_jupyter_directory(self, user_dir: Path):
...
+        Ensure the user's virtual environment exists. If it does not exist, it is
+        created with the system site-packages included.
+        """
+        if not user_env_dir.exists():
+            user_env_dir.mkdir(parents=True)
+            self.log.info(f'Creating virtual environment for {self.user.name}')
+            try:
+                # Create a virtual environment with system site-packages access
+                subprocess.run(['python3', '-m', 'venv', str(user_env_dir), '--system-site-packages'],
+                               check=True)
```
MrCreosote, line 149: https://docs.python.org/3/library/venv.html#venv.EnvBuilder.create might be better here

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -224,3 +232,88 @@ def test_ensure_user_jupyter_directory_user_dir_does_not_exist():
...
+@patch.dict(os.environ, {
+    'PATH': '/usr/local/bin:/usr/bin:/bin',
+    'JUPYTERHUB_CONFIG_DIR': '/etc/jupyterhub',
+})
+def test_configure_environment_missing_pythonpath(spawner):
+    user_dir = Path('/home/testuser')
+    user_env_dir = Path('/home/testuser/.venv')
+    username = 'testuser'
+
+    spawner._configure_environment(user_dir, user_env_dir, username)
```
MrCreosote, line 316: I'm confused, reading the code I'd expect this to throw a keyerror

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -224,3 +232,88 @@ def test_ensure_user_jupyter_directory_user_dir_does_not_exist():
...
+
+    # Check that PYTHONPATH is set correctly even when it's not in the original environment
+    assert f"{user_env_dir}/lib/python3.11/site-packages:" in spawner.environment['PYTHONPATH']
```
MrCreosote, line 319: Will the colon cause problems down the line if any other code tries to append stuff to the path?

src/jupyterhub_config/custom_spawner.py
```diff
@@ -126,4 +133,38 @@ def _ensure_user_jupyter_directory(self, user_dir: Path):
...
+            except subprocess.CalledProcessError as e:
+                self.log.error(f'Failed to create virtual environment: {e}')
+                raise
```
MrCreosote, line 152: Did you mean to leave this untested?

Tianhao-Gu: added test.

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: It should. My IDE (Pycharm) auto appended some pythonpath so that it can run tests.

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: It's the spawner environment variable (user's individual environment). I think it's good to copy os environment to individual user's. Maybe we need a white list in the future to hide some os env from users.

Tianhao-Gu: It's the spawner environment variable (user's individual environment). I think it's good to copy os environment to individual user's. Maybe we need a white list in the future to hide some os env from users.

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: I'm guessing GHA also has a PYTHONPATH

src/jupyterhub_config/custom_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_spawner.py
MrCreosote: I guess I'm wondering why the spawner doesn't populate it

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: Anyway, moot now :+1:

Tianhao-Gu: yea yea. I think so.

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: 👍

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: I think it should be fine, `xxx/python3.11/site-packages::user_defined_path` should work. But just in case, I added a statement to check os.environ['PYTHONPATH']

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: like automatically populate it? I am not sure. Maybe by default, they don't want to overwrite the os.env. There maybe a flag setting I haven't figured out yet.

MrCreosote: > Maybe by default, they don't want to overwrite the os.env.

I'm confused now - I thought self.environment was separate from os.env. It seems like you're overwriting self.environment with os.env. My basic assumption is that if that was the right thing to do the Jupyter devs would've done it, but maybe I'm wrong.

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: :+1:

src/jupyterhub_config/custom_spawner.py
MrCreosote: :+1:

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: Although now the no pythonpath case presumably isn't tested

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: yea. That's exactly what I am trying to do here, copy missing os.env to user's spawner env. It has things like `YARN_CONF_DIR` and `SPARK_MASTER_URL` that we need. But it's not automatically set to the user's environment if I don't explicitly set them.

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: 👍 updated test for the missing pythonpath case.

src/jupyterhub_config/custom_spawner.py
MrCreosote: Ok. I would probably use an allow list here rather than dumping everything in there but I leave that up to you

test/src/jupyterhub_config/custom_spawner_test.py
MrCreosote: :+1:

MrCreosote (approved):

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: 👍 I will dump everything for now just in case I might miss something.

PR #83, opened 2024-09-11
add local auth

Tianhao-Gu:


Pipfile
```diff
@@ -4,16 +4,17 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-jupyterlab= "==4.2.4"
+jupyterlab= "==4.2.5"
 jupyterhub = "==5.1.0"
-oauthenticator = "==16.3.1"
+jupyterhub-nativeauthenticator = "==1.2.0"
+oauthenticator = "==17.0.0"
```
MrCreosote, line 10: Both of these are needed?

config/jupyterhub_config.py
```diff
@@ -4,24 +4,40 @@
...
-# TODO: Change the authenticator class to a secure one (e.g. GitHubOAuthenticator)
-c.JupyterHub.authenticator_class = 'jupyterhub.auth.DummyAuthenticator'
-c.Authenticator.allowed_users = {'spark_user', 'test_user1', 'test_user2'}
-c.DummyAuthenticator.password = os.environ['JUPYTERHUB_ADMIN_PASSWORD']
+# Set the authenticator class to nativeauthenticator
+# ref: https://native-authenticator.readthedocs.io/en/latest/quickstart.html
+c.JupyterHub.authenticator_class = 'native'
+c.JupyterHub.template_paths = [f"{os.path.dirname(nativeauthenticator.__file__)}/templates/"]
+c.NativeAuthenticator.open_signup = True
+c.NativeAuthenticator.check_common_password = True
```
MrCreosote, line 21: what do these mean?

config/jupyterhub_config.py
```diff
@@ -4,24 +4,40 @@
...
+c.NativeAuthenticator.minimum_password_length = 6
```
MrCreosote, line 22: Way too short in general, but for a stopgap I guess it doesn't matter

config/jupyterhub_config.py
```diff
@@ -4,24 +4,40 @@
...
 
+# Set up the admin user
 c.Authenticator.admin_users = {'spark_user'}
+# TODO set admin user password to os.environ['JUPYTERHUB_ADMIN_PASSWORD'] automatically - currently spark_user is created manually with the signup page
```
MrCreosote, line 26: Is there anything preventing a user from creating their name as spark_user other than an admin doing it first via the uI?

config/jupyterhub_config.py
```diff
@@ -4,24 +4,40 @@
...
+c.Authenticator.allow_all = True
```
MrCreosote, line 27: What does this do?

config/jupyterhub_config.py
```diff
@@ -4,24 +4,40 @@
...
+
+c.JupyterHub.cookie_secret_file = f"{os.environ['JUPYTERHUB_USER_HOME']}/jupyterhub_cookie_secret"
```
MrCreosote, line 29: What is this for?

config/jupyterhub_config.py
```diff
@@ -4,24 +4,40 @@
...
+c.JupyterHub.db_url = f"sqlite:///{os.environ['JUPYTERHUB_USER_HOME']}/jupyterhub.sqlite"
```
MrCreosote, line 30: If this is the user database it probably shouldn't be stored in the user home dir

config/jupyterhub_config.py
```diff
@@ -4,24 +4,40 @@
...
 
 c.JupyterHub.spawner_class = VirtualEnvSpawner
 
+# Create a group to indicate users with read/write access to MinIO
+c.JupyterHub.load_groups = {
+    'minio_rw': [],
+}
```
MrCreosote, line 37: How do users get added to this group? I'm not sure how this actually works, come to think of it - the creds are in the env so the users by default have the creds right?

Pipfile
Tianhao-Gu: just `jupyterhub-nativeauthenticator` for now. But I thought it's better to get pipfile fixed in one shot as we might use oauth.

config/jupyterhub_config.py
Tianhao-Gu: those variables are described in the ref link commented above.

config/jupyterhub_config.py
Tianhao-Gu: they cannot. The auth class checks current db to make sure there is no existing user.

config/jupyterhub_config.py
Tianhao-Gu: add ref for that

config/jupyterhub_config.py
Tianhao-Gu: keep cookies I guess.

config/jupyterhub_config.py
Tianhao-Gu: where would we store it?

Pipfile
MrCreosote: I wouldn't add a dependency that's not currently needed

config/jupyterhub_config.py
MrCreosote: Neither of those are in the linked page AFAICT

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
MrCreosote: Should it be in the user home dir if it's holding secrets, then?

config/jupyterhub_config.py
Tianhao-Gu: you can do `'minio_rw': ['user_name1', 'user_name2'],` here or calling jupyterhub API to add users which is what I am planning to do. 

> users by default have the creds right

hmmm no. Users need to setup their own password when they signup an account.

MrCreosote: I mean the Minio creds. How would this group stop them from getting the Minio creds from the env?

Pipfile
Tianhao-Gu: 👍

MrCreosote: :+1:

config/jupyterhub_config.py
Tianhao-Gu: 👍 added a ref link for that. You just need to click on `Optional Configuration` from the main link.

config/jupyterhub_config.py
Tianhao-Gu: changed to 8.

config/jupyterhub_config.py
Tianhao-Gu: they don't. We create users if they don't exist at startup.

config/jupyterhub_config.py
Tianhao-Gu: in next PR, I will set minio creds accordingly based on user group.

config/jupyterhub_config.py
MrCreosote: I mean a decent password should be >16 or so characters, 8 is going to be in rainbow tables, but like I say if this is temporary and not exposed to the network is probably doesn't matter.

config/jupyterhub_config.py
Tianhao-Gu: I think that would be the same, volume is mounted, whether it's `/jupyterhubsecrets` or `/jupyterhub/users_home`. Users know where to find it, users can see it. But no r/w access.

config/jupyterhub_config.py
MrCreosote: Ok, but the admin users are created before any users have a chance to create an account?

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
Tianhao-Gu: no. It just happened we already created `spark_user` in Dockerfile. If the admin user doesn't exist, it will go through user creation code. The username must be the same as what's listed in `c.Authenticator.admin_users` to be an admin.

config/jupyterhub_config.py
MrCreosote: wait users don't have r/w to JUPYTERHUB_USER_HOME ?

config/jupyterhub_config.py
Tianhao-Gu: It's a global file. Not created individually for each user.

config/jupyterhub_config.py
Tianhao-Gu: no. they have w/r to their own dir. But not `JUPYTERHUB_USER_HOME`

config/jupyterhub_config.py
MrCreosote: So theoretically, if user X came along fast after the container came up, they could create an account called `spark_user` an be an admin?

config/jupyterhub_config.py
Tianhao-Gu: yea. it will be set via by spawner `self.environment`.

config/jupyterhub_config.py
MrCreosote: That's why I'm asking why it should be in the home dir at all

config/jupyterhub_config.py
MrCreosote: Isn't that the root of their home dirs?

config/jupyterhub_config.py
Tianhao-Gu: no, I don't think so. `spark_user` is already created by dockerfile.

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
Tianhao-Gu: It's just a convenient to put all jupyterhub related files.

config/jupyterhub_config.py
MrCreosote: But the juypterhub auth system doesn't have an account set up for that yet. So what prevents someone other than you (or whoever) from taking it?

config/jupyterhub_config.py
MrCreosote: I would definitely not put password / secret information in the user directory hierarchy

config/jupyterhub_config.py
Tianhao-Gu: It is. But that dir belongs to `spark_user`

config/jupyterhub_config.py
Tianhao-Gu: taking what? I am completely lost here. What's the concern here? Someone will create a `spark_user`? Before the first time the user_db gets created?

MrCreosote: The concern is that the hub container starts and some user comes along and creates a `spark_user` account, and thus gets admin rights. It's not clear to me whether that can happen or not, and if not, why not

config/jupyterhub_config.py
Tianhao-Gu: 👍

config/jupyterhub_config.py
Tianhao-Gu: this will also be in `jupyterhub_secrets` dir

config/jupyterhub_config.py
Tianhao-Gu: yea. It can happen before we manually created the user. I don't see it's a concern or something likely to happen. I have a todo item to set the admin password. Also once the admin user is created, it won't happen. So theoretically,  it can happen, I don't see a way to prevent it. I can move locally created user db to the mounted dir so that auth read from existing db.

MrCreosote: If you want to leave it as is for now but create an issue that's fine with me

>  I can move locally created user db to the mounted dir so that auth read from existing db.

Don't you need to do this to preserve user account over container recreations?

Tianhao-Gu: no. Users will signup by themselves.

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
MrCreosote: I don't understand how this answer pertains to the question

Tianhao-Gu: > Don't you need to do this to preserve user accounts over container recreations?

Users will signup and jupyterhub service will create an account accordingly in `jupyterhub.sqlite` and spawner will create a system user accordingly. So I don't need to move locally created db to the mounted dir.

MrCreosote: But if the db isn't in a mounted dir, everytime the container is recreated all the user accounts will be wiped

Tianhao-Gu: yea. db is in a mounted dir. I was saying I can copy the locally created db which has spark_user records to the mounted dir before the first time container starts. That would prevent some random user create spark_user.

MrCreosote: Ok, now I'm confused as to why there are 2 db locations, the mounted location and the 'locally created' location. Shouldn't there be just one location?

Tianhao-Gu: no. In my local environment (on my laptop), I have created a db for testing purposes. And that db has `spark_user` records. I was saying I could copy that db to the mounted location in the server, so that before container starts, the db in the mounted location exists.

MrCreosote: Oh, you add a prefilled db with the admin accounts to the container, I see.

Ok. If you want to do that that sounds reasonable (unless there's some drawback like blowing away user settings that are added to the db later from the UI), but I don't think it's required since this is a temporary solution and presumably won't be doing to prod.

MrCreosote (approved):

PR #88, opened 2024-09-13
fix docker volumn mount + notebook dir for admin

Tianhao-Gu:


src/jupyterhub_config/custom_spawner.py
```diff
@@ -206,7 +206,8 @@ def _configure_notebook_dir(self, username: str, user_dir: Path):
         """
         if self.user.admin:
             self.log.info(f'Admin user detected: {username}. Setting up admin workspace.')
-            self.notebook_dir = '/cdm_shared_workspace'
+            # global users' home directory
+            self.notebook_dir = str(user_dir.parent)
```
MrCreosote, line 210: So basically admin notebooks are all shared and live in /home and user notebooks live in /home/{user}? And users can read /home/{user} but not home?

test/src/jupyterhub_config/custom_spawner_test.py
```diff
@@ -446,17 +446,19 @@ def test_configure_environment_missing_pythonpath(spawner):
...
-    (True, '/cdm_shared_workspace'),  # Admin user case
-    (False, 'to_be_defined_in_the_test')  # Non-admin user case
+@pytest.mark.parametrize("is_admin", [
+    True,  # Admin user case
+    False  # Non-admin user case
 ])
-def test_configure_notebook_dir(is_admin, expected_dir, spawner, caplog):
+def test_configure_notebook_dir(is_admin, spawner, caplog):
     spawner.user.admin = is_admin
     username = 'testuser'
```
MrCreosote, line 455: This would probably be simpler as just a loop at this point, like

```
for b in [True, False]:
    spawner.user.admin = b
    ...
```

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: I think `str('/')` makes more sense for admin user.

MrCreosote: or /root

MrCreosote: or /admin

Tianhao-Gu: 👍 just changed it to '/' for now. I think most majority users won't be admin.

test/src/jupyterhub_config/custom_spawner_test.py
Tianhao-Gu: 👍

src/jupyterhub_config/custom_spawner.py
MrCreosote: That means any notebooks they create won't be persisted, I assume

MrCreosote (approved):

src/jupyterhub_config/custom_spawner.py
Tianhao-Gu: yea yea. I think they (or most likely just us) just need to remember to create notebook in the `jupyterhub/users_home` dir.

PR #92, opened 2024-09-19
update lib and dockerfile for docker spawner

Tianhao-Gu:


Dockerfile
```diff
@@ -98,16 +98,28 @@ ENV CDM_SHARED_DIR=/cdm_shared_workspace
...
+
 # Allow spark_user to use sudo without a password
 # TODO: use `sudospawner` in JupyterHub to avoid this (https://jupyterhub.readthedocs.io/en/stable/howto/configuration/config-sudo.html)
 RUN echo "spark_user ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 
 # Switch back to non-root user
-USER spark_user
+# Facing permission errors when accessing the Docker API as a non-root user
+#USER spark_user
+
```
MrCreosote, line 123: So is this dockerfile for the spawner, and hence running at root?

Dockerfile
```diff
@@ -98,16 +98,28 @@ ENV CDM_SHARED_DIR=/cdm_shared_workspace
...
 # Set a directory for hosting Jupyterhub db and cookie secret
 ENV JUPYTERHUB_SECRETS_DIR=/jupyterhub_secrets
 RUN mkdir -p ${JUPYTERHUB_SECRETS_DIR}
 RUN chown -R spark_user:spark ${JUPYTERHUB_SECRETS_DIR}
 
+# Set a directory for hosting postgres db data - used by Spark to store Delta tables
+ENV JUPYTERHUB_POSTGRES_DIR=/cdm-postgres
+RUN mkdir -p ${JUPYTERHUB_POSTGRES_DIR}
+RUN chown -R spark_user:spark ${JUPYTERHUB_POSTGRES_DIR}
+
```
MrCreosote, line 115: I thought we were using a remote postgres instance for hive. What is this for?

Dockerfile
```diff
@@ -98,16 +98,28 @@ ENV CDM_SHARED_DIR=/cdm_shared_workspace
 RUN mkdir -p ${CDM_SHARED_DIR} && chmod -R 777 ${CDM_SHARED_DIR}
 RUN chown -R spark_user:spark $CDM_SHARED_DIR
 
+# Set a directory for hosting Hive metastore files - defined in config/hive-site-template.xml
+ENV HIVE_METASTORE_DIR=$CDM_SHARED_DIR/hive_metastore
+RUN mkdir -p ${HIVE_METASTORE_DIR}
+RUN chown -R spark_user:spark $HIVE_METASTORE_DIR
```
MrCreosote, line 104: Presumably this needs to be writable by ppl who can add deltatables?

Dockerfile
Tianhao-Gu: hmmm I currently use the same dockerfile (image) for both jupyterhub server and the spawner (user). The hub needs to run docker socket so that it can spin up containers for users. That's where I encounter permission issues with spark_user. Boris actually suggests that we use a simpler image for the server. And use this docker file specifically for the user. I think that makes sense to do. But maybe later work.

Dockerfile
Tianhao-Gu: Yes we are. I am getting confused here too. Do users need to read data from postgres at all? If not, I don't think we need this. I am planning to mount `/cdr/cdm/jupyter/cdm-postgres:/cdm-postgres`. Maybe that's not needed?

Dockerfile
MrCreosote: Yeah, I would def not run the user containers as root, that makes sense. Later work is fine

Dockerfile
Tianhao-Gu: Yea. yea. That's right.

Dockerfile
MrCreosote: I assume they need to read / write to the remote postgres as that IIUC, stores the table metadata for the deltalake tables. I don't understand why we'd need local postgres tables at all

Dockerfile
MrCreosote: So will this break that?

Dockerfile
Tianhao-Gu: no. it's not local postgres. But using the remote ones. The remote postgres does `/cdr/cdm/jupyter/cdm-postgres:/var/lib/postgresql/data` in Rancher.

Dockerfile
Tianhao-Gu: I don't think so. I will handle permissions before spawner starts. Same as local spawner we did before.

MrCreosote: :+1:

Dockerfile
Tianhao-Gu: 👍 just want to get something that works for now. Will revisit.

Dockerfile
MrCreosote: If it's remote why does it need a local folder at all?

Tianhao-Gu: 👍 yea. Now I am thinking. You are right. Removed.

MrCreosote (approved):

PR #93, opened 2024-09-19
customer docker spawner class

Tianhao-Gu:
Copied directly from `custom_spawner.py` which will be deleted once docker spawner is setup.

### DIFF
--- a/src/jupyterhub_config/custom_docker_spawner.py
+++ b/src/jupyterhub_config/custom_docker_spawner.py
@@ -1,0 +1,133 @@
+import os
+import venv
+from pathlib import Path
+
+from dockerspawner import DockerSpawner
+
+
+class CustomDockerSpawner(DockerSpawner):
+    RW_MINIO_GROUP = 'minio_rw'
+
+    def start(self):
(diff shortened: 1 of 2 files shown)


src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -0,0 +1,133 @@
+import os
+import venv
+from pathlib import Path
+
+from dockerspawner import DockerSpawner
+
+
+class CustomDockerSpawner(DockerSpawner):
```
Tianhao-Gu, line 8: it now inherits from the DockerSpawner module. All other funcs are copied from `custom_spawner`

MrCreosote:
To be clear I've already reviewed everything in the file here; the only change is the superclass?

Tianhao-Gu:
> To be clear I've already reviewed everything in the file here; the only change is the superclass?

yea. Directly copied and pasted from  previous `custom_spawner` class.

MrCreosote (approved):
Approving w/o further review based on the above

PR #95, opened 2024-09-23
start jupyterhub-singleuser within docker spawer

Tianhao-Gu:


src/jupyterhub_config/hub_singleuser.py
```diff
@@ -0,0 +1,143 @@
...
+    def setup_environment(self):
+        """
+        Set up the user's environment by ensuring the system user exists, and the workspace
+        permissions are correctly configured.
+        """
+        self.log.info(f"Setting up environment for user: {self.username}")
+        self._ensure_system_user()
+        self._ensure_workspace_permission()
+
+    def start_single_user_server(self):
```
Tianhao-Gu, line 43: copied from [customer_spawner.py](https://github.com/kbase/cdm-jupyterhub/blob/main/src/jupyterhub_config/custom_spawner.py#L54-L58)

src/jupyterhub_config/hub_singleuser.py
```diff
@@ -0,0 +1,143 @@
...
+        self.log.info(f'Starting single-user server for {self.username} with environment: {self.environment}')
+        env_vars = [f'{key}={value}' for key, value in self.environment.items()]
+
+        cmd = ['sudo', '-E', '-u', self.username, 'env'] + env_vars + [
+            os.path.join(os.environ['JUPYTERHUB_CONFIG_DIR'], 'spawn_notebook.sh')]
+
+        self.log.info(f'Executing command: {" ".join(cmd)}')
+        subprocess.run(cmd, check=True)
+
+    def _ensure_system_user(self):
```
Tianhao-Gu, line 56: mostly copied from [customer_spawner.py](https://github.com/kbase/cdm-jupyterhub/blob/main/src/jupyterhub_config/custom_spawner.py#L62). But using `pwd` and `grp` to check existing user and group.

src/jupyterhub_config/hub_singleuser.py
```diff
@@ -0,0 +1,143 @@
...
+                    self.log.info(f'Creating system user: {self.username}')
+                    subprocess.run(useradd_cmd, check=True)
+
+            except subprocess.CalledProcessError as e:
+                raise ValueError(f'Failed to create system user: {e}')
+
+            finally:
+                fcntl.flock(lock, fcntl.LOCK_UN)
+
+    def _ensure_workspace_permission(self):
```
Tianhao-Gu, line 101: copied from [custom_spawner.py](https://github.com/kbase/cdm-jupyterhub/blob/main/src/jupyterhub_config/custom_spawner.py#L202)

src/jupyterhub_config/hub_singleuser.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        Initializes the environment manager for the specified user.
+        :param username: The name of the user for whom the environment is set up.
+        :param groupname: The name of the OS system group to which the user belongs.
+        """
+        self.username = username  # OS system username
+        self.groupname = groupname  # OS system group
+
+        # Inherit the current environment variables
+        self.environment = dict(os.environ)
```
MrCreosote, line 28: This still makes me nervous...

src/jupyterhub_config/hub_singleuser.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def __init__(self,
+                 username,
+                 groupname: str = 'jupyterhub', ):
+        """
+        Initializes the environment manager for the specified user.
+        :param username: The name of the user for whom the environment is set up.
+        :param groupname: The name of the OS system group to which the user belongs.
+        """
+        self.username = username  # OS system username
```
MrCreosote, line 24: I don't understand why the juipyterhub username and the linux username need to be the same

src/jupyterhub_config/hub_singleuser.py
```diff
@@ -0,0 +1,143 @@
...
+        try:
+            user_info = pwd.getpwnam(self.username)
+        except KeyError:
+            raise ValueError(f'System user {self.username} does not exist')
+        gid = user_info.pw_gid
+        group_name = grp.getgrgid(gid).gr_name
+
+        self.log.info(f'Configuring workspace permissions for {self.username}')
+        subprocess.run(['sudo', 'chown', '-R', f'{self.username}:{group_name}', self.user_dir], check=True)
+        subprocess.run(['sudo', 'chmod', '-R', '777', self.user_dir], check=True)
```
MrCreosote, line 114: Is this still needed or is 700 or 770 ok?

Also this is going to make all files executable, which I assume you don't want

Tianhao-Gu: 👍 changed to 770.

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: somewhere down the line (I think it's auth) will need to check that system user exist with the same jupyterhub username.

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: I will get a whitelist later once system is up and running.

src/jupyterhub_config/hub_singleuser.py
MrCreosote: I don't quite get that - why would the linux username need to match the auth username?

src/jupyterhub_config/hub_singleuser.py
MrCreosote: Making all the files executable?

src/jupyterhub_config/hub_singleuser.py
MrCreosote: :+1:

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: maybe for some spawner it's not required. But I don't see it concerns us.

Tianhao-Gu: https://discourse.jupyter.org/t/is-authentication-without-system-users-possible/1770

MrCreosote: > But I don't see it concerns us.

If the system user is the same for all containers, then we don't have to worry about the YARN queue stuff

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: 👍 changed to 750. I am really still trying to make it work first, before considering security, etc.

MrCreosote: 750 still makes the files executable for the owner but doesn't let the group traverse the directories. You'd probably need to do it in 2 steps, one handling directories and one handling files, unless there's some build in way in chmod to do that.

That said if you want to make an issue and move on that's fine

Tianhao-Gu: I didn't get the concern here. Why we shouldn't make all files executable for the owner?

MrCreosote: Because most files aren't executable and shouldn't be executed.

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: but then how are files shared with the same system user and different juptyerhub user?

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: but the workspace and container is user's. I don't see why not provide max user's need.

src/jupyterhub_config/hub_singleuser.py
MrCreosote: Well, I don't know how jupyterhub works, but my assumption is that it would be like a lot of other software and the system user wouldn't matter. The software would run as the system user but the username associated with the software isn't an existing linux account at all.

src/jupyterhub_config/hub_singleuser.py
MrCreosote: There is no need to make non-executable files executable

https://security.stackexchange.com/questions/192197/reason-to-not-use-chmod-r-777-on-internal-server-for-project-source-code

Tianhao-Gu: But it also doesn't hurt. And really not our concerns now.

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: maybe there are workarounds. Is it worth the time to explore it, absolutely no. Yarn is working fine with different users.

MrCreosote: > Yarn is working fine with different users.

Not if it's bypassing fair share it isn't

src/jupyterhub_config/hub_singleuser.py
MrCreosote: > But it also doesn't hurt.

It does hurt for the reasons in the link and general security principles. But as I said

> That said if you want to make an issue and move on that's fine

Tianhao-Gu: I really don't know what to put on the issue. Do you think you can make one? I just don't see any security concerns or principles being violated by assigning users 750 permissions on their own workspace within their isolated container.

MrCreosote: The security concerns are listed above in the SO post

For the ticket, just something like "Don't chmod non-executable files to executable" and link this PR, I'd say

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: I will explore some yarn setting (`spark.yarn.queue`) on the client's side.

MrCreosote: If you just use the same user for the containers you don't have to worry about this

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: 👍 https://github.com/kbase/cdm-jupyterhub/issues/96

src/jupyterhub_config/hub_singleuser.py
Tianhao-Gu: I created another issue for this. https://github.com/kbase/cdm-jupyterhub/issues/97

It's really not how Jupyterhub works as far as I understand.

MrCreosote: If you're saying that JH requires the linux username to match the JH username then yeah, that won't work and we'll have to look into modifying yarn or how the submit works

MrCreosote (approved):

PR #98, opened 2024-09-24
update jupterhub config to enable docker spanwer

Tianhao-Gu:


config/jupyterhub_config.py
```diff
@@ -33,13 +33,32 @@
...
+c.DockerSpawner.cpu_limit = 4
+c.DockerSpawner.mem_limit = '16G'
+
+# The network name that Docker containers will use to communicate
+network_name = os.environ.get('NETWORK_NAME')
+if network_name:
+    c.DockerSpawner.network_name = network_name
+c.DockerSpawner.use_internal_ip = True
+environment = os.environ.get('ENVIRONMENT', 'prod').lower()
+c.DockerSpawner.remove = environment != 'dev'
```
MrCreosote, line 59: Can you add a comment describing the purpose of this line?

docker-compose.yaml
```diff
@@ -210,10 +228,16 @@ services:
       - POSTGRES_DB=hive
       - POSTGRES_URL=postgres:5432
       - JUPYTERHUB_ADMIN_PASSWORD=testpassword123
+      - NETWORK_NAME=cdm-jupyterhub-network
+      - JUPYTERHUB_USER_IMAGE=cdm-jupyterhub-cdm_jupyterhub:latest
```
MrCreosote, line 232: You don't want to pin this?

MrCreosote, line 232: Is there eventually going to be a small spawner image and a user image or are you thinking of just having the 1

docker-compose.yaml
```diff
@@ -232,4 +256,11 @@ services:
...
-      - ./cdr/cdm/jupyter/cdm-postgres:/var/lib/postgresql/data  # For local development only. In Rancher development, PostgreSQL data shouldn't be stored in a shared mount.
\ No newline at end of file
+      - ./cdr/cdm/jupyter/cdm-postgres:/var/lib/postgresql/data  # For local development only. In Rancher development, PostgreSQL data shouldn't be stored in a shared mount.
+    networks:
+      - cdm-jupyterhub-network
+
+networks:
+  # Created by `docker network create cdm-jupyterhub-network`
+  cdm-jupyterhub-network:
+    external: true
```
MrCreosote, line 266: What does external do here?

docker-compose.yaml
```diff
@@ -210,10 +228,16 @@ services:
       - POSTGRES_DB=hive
       - POSTGRES_URL=postgres:5432
       - JUPYTERHUB_ADMIN_PASSWORD=testpassword123
+      - NETWORK_NAME=cdm-jupyterhub-network
+      - JUPYTERHUB_USER_IMAGE=cdm-jupyterhub-cdm_jupyterhub:latest
+      - JUPYTERHUB_MOUNT_BASE_DIR=/cdr/cdm/jupyter
```
MrCreosote, line 233: This doesn't look like it's used in this PR

scripts/notebook_entrypoint.sh
```diff
@@ -31,6 +31,12 @@ elif [ "$JUPYTER_MODE" = "jupyterhub" ]; then
   unset NOTEBOOK_DIR
 
   jupyterhub -f "$JUPYTERHUB_CONFIG_DIR"/jupyterhub_config.py
+elif [ "$JUPYTER_MODE" = "jupyterhub-singleuser" ]; then
+  echo "Starting Jupyter Notebook for user: $JUPYTERHUB_USER"
+  unset NOTEBOOK_DIR
+
+  python /src/jupyterhub_config/hub_singleuser.py
+
```
MrCreosote, line 39: There's going to be a lot of jupyterlab and jupyterhub multiuser cruft to clean up at some point...

MrCreosote (commented):
This is pretty hard for me to completely understand the whole picture, but I don't see any huge issues

config/jupyterhub_config.py
Tianhao-Gu: 👍

docker-compose.yaml
Tianhao-Gu: in Rancher, it's pined. But locally latest is the one we want to test against.

Tianhao-Gu: spawner image will be this one since it has all the packages users need. Maybe having a light image for the jupterhub server.

docker-compose.yaml
Tianhao-Gu: it indicates the network is created by `docker network create`. I wasn't able to get it working locally with docker compose created networks.

docker-compose.yaml
Tianhao-Gu: yea not in this PR. But used here. https://github.com/kbase/cdm-jupyterhub/blob/main/src/jupyterhub_config/custom_docker_spawner.py#L150

config/jupyterhub_config.py
MrCreosote: Under what conditions does the container get removed?

docker-compose.yaml
MrCreosote: Why wasn't this line needed then?

docker-compose.yaml
MrCreosote: :+1:

MrCreosote: I think we mean different things by "spawner". When I say spawner I'm talking about whatever is responsible for spawning the user containers

docker-compose.yaml
MrCreosote: What implications does that have?

Do you have to do something similar in Rancher?

config/jupyterhub_config.py
Tianhao-Gu: when the container stops (error, user can also manually restart spawner).

docker-compose.yaml
Tianhao-Gu: yea 
> Maybe having a light image for the jupterhub server..

docker-compose.yaml
Tianhao-Gu: I am working with Devops on this.

docker-compose.yaml
Tianhao-Gu: to keep PR small, otherwise PR review takes very long.

config/jupyterhub_config.py
MrCreosote: > user can also manually restart spawner

Here you're talking about the user image right? Not the user image spawner? 

How would the user restart the spawner?

Can you update the comment to explain when the image is removed?

docker-compose.yaml
MrCreosote: Ok, so it's not clear what's going to happen in rancher yet.

I still don't understand what the difference is between the two network modes in docker compose

config/jupyterhub_config.py
Tianhao-Gu: 👍

docker-compose.yaml
Tianhao-Gu: With the one docker compose created automatically, I couldn't get jupterhub working locally. Hostname is not resolved correctly which is what we see in Rancher now. I don't know I have the best answer for this question. Still trying to figure it out. But with this setting, at least I can get jupterhub working properly locally with docker compose.

config/jupyterhub_config.py
Tianhao-Gu: >Here you're talking about the user image right? Not the user image spawner?

yea the user image. Not jupyterhub server. 

> How would the user restart the spawner?
On the home page, users can stop and restart server. Similar to NERSC. 

Screenshot 2024-09-25 at 2 05 18 PM

docker-compose.yaml
MrCreosote: Ok, sounds like this is "well it works but I don't know why" situation.

The reason I was asking this is because I don't know how that would translate to Rancher and if it'd have any security implications

MrCreosote (approved):

docker-compose.yaml
Tianhao-Gu: yea. I think I will try to understand and how it works in Rancher and maybe come back and update docker compose to make it similar if not the same as Rancher.

PR #101, opened 2024-09-30
add favorite feature

Tianhao-Gu:


src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -167,5 +172,55 @@ def _ensure_user_volume(self):
...
+        Configure the JupyterLab favorites for the user.
+        """
+        self.log.info('Configuring JupyterLab favorites for user')
+
+        if not favorites:
+            favorites = [user_dir]
+
+        # Ensure the user's home directory is always in the favorites
+        if user_dir not in favorites:
+            favorites.append(user_dir)
```
MrCreosote, line 190: Why not just use a `set`?

Then you could do:

```
favorites = {user_dir} if not favorites else favorites | {user_dir}

File: src/jupyterhub_config/custom_docker_spawner.py [hunk 2]
```diff
@@ -167,5 +172,55 @@ def _ensure_user_volume(self):
...
+
+        # Path to the JupyterLab favorites configuration file
+        jupyterlab_favorites_path = user_dir / '.jupyter' / 'lab' / 'user-settings' / '@jlab-enhanced' / 'favorites' / 'favorites.jupyterlab-settings'
+        favorites_dir = jupyterlab_favorites_path.parent
+
+        if not favorites_dir.exists():
+            favorites_dir.mkdir(parents=True, exist_ok=True)
```
MrCreosote, line 197: No need for the if really, you can just run this every time

src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -167,5 +172,55 @@ def _ensure_user_volume(self):
...
+
+        if jupyterlab_favorites_path.exists():
+            with open(jupyterlab_favorites_path, 'r') as f:
```
MrCreosote, line 200: This is a race condition - if this method is run concurrently one can clobber the other's favorites

src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -167,5 +172,55 @@ def _ensure_user_volume(self):
...
+
+        existing_fav_set = {(fav["root"], fav["path"]) for fav in exist_favorites.get('favorites', [])}
+
+        for fav in favorites:
+
+            if not fav.is_dir():
+                raise ValueError(f"Favorite {fav} is not a directory or does not exist")
+
+            root_str = str(fav)
+            path_str = ""
```
MrCreosote, line 214: What is the purpose of path_str if it's always the empty string?

src/jupyterhub_config/custom_docker_spawner.py
Tianhao-Gu: The `jupyterlab_favorites_path` is specific to each user + the `user_dir` is mounted exclusively for the individual user, so I don't see any way for others to access or read someone else's `jupyterlab_favorites_path`.

src/jupyterhub_config/custom_docker_spawner.py
Tianhao-Gu: Without it, the extension doesn't function properly. I assume it doesn't set default values internally. I couldn't find any documentation on this. I observed how NERSC handles it and replicated that behavior here.

src/jupyterhub_config/custom_docker_spawner.py

src/jupyterhub_config/custom_docker_spawner.py

src/jupyterhub_config/custom_docker_spawner.py
MrCreosote: It doesn't have to be another person, it could be the same user trying to add multiple favorites in quick succession

src/jupyterhub_config/custom_docker_spawner.py
MrCreosote: So it's always an empty string @ NERSC too?

src/jupyterhub_config/custom_docker_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_docker_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_docker_spawner.py
Tianhao-Gu: they did it the other way. I don't think it matters. 
```
(base) tgu@login34 favorites  $ cat favorites.jupyterlab-settings 
{
    // Favorites
    // @jlab-enhanced/favorites:favorites
    // Favorites settings.
    // **********************************

    // Favorites
    // The list of favorites.
    "favorites": [
        {
            "root": "/",
            "path": "global/homes/t/tgu",
            "name": "$HOME",
            "contentType": "directory",
            "iconClass": "jp-MaterialIcon jp-FolderIcon",
            "default": true,
            "hidden": false
        },
        {
            "root": "/",
            "path": "global/cscratch1/sd/tgu",
            "name": "$CSCRATCH",
            "contentType": "directory",
            "iconClass": "jp-MaterialIcon jp-FolderIcon",
            "default": true,
            "hidden": true
        },
        {
            "root": "/",
            "path": "pscratch/sd/t/tgu",
            "name": "$PSCRATCH",
            "contentType": "directory",
            "iconClass": "jp-MaterialIcon jp-FolderIcon",
            "default": true
        },
        {
            "root": "/",
            "contentType": "directory",
            "iconLabel": "ui-components:folder",
            "path": "global/cfs/cdirs/kbase/collections"
        }
    ]
}
```

Tianhao-Gu: yea. I am guessing the extension is using `os.path.join(root, path)` internally or something like that.

src/jupyterhub_config/custom_docker_spawner.py

MrCreosote (approved):

PR #102, opened 2024-10-01
add shared dir to user's favorite list

Tianhao-Gu:


Dockerfile
```diff
@@ -98,6 +98,11 @@ ENV CDM_SHARED_DIR=/cdm_shared_workspace
 RUN mkdir -p ${CDM_SHARED_DIR} && chmod -R 777 ${CDM_SHARED_DIR}
 RUN chown -R spark_user:spark $CDM_SHARED_DIR
 
+# TODO: Config through a config file or DB as the number of groups increases.
```
MrCreosote, line 101: How are you thinking this is going to work? Is there going to be a mount per group?

MrCreosote (approved):

Dockerfile
Tianhao-Gu: I am thinking in long-term plan, we will first get a list of dirs that are shared to the login user. And then mount those dir to that user.

PR #107, opened 2024-10-16
add bashrc ext files

Tianhao-Gu:


MrCreosote:
Why the .ext file extensions? Is it something to do with this? https://docs.nersc.gov/environment/#dotfiles-and-customizing-your-environment

Tianhao-Gu:
> Why the .ext file extensions? Is it something to do with this? https://docs.nersc.gov/environment/#dotfiles-and-customizing-your-environment

yea. I copied the implementation idea from NERSC. But we could change the `.ext` to something our own. Our use case currently is much simpler. I will basically just use them as template files and create .bashrc and .bash_profile in users' home dir.

MrCreosote:
Why have extensions at all then?

Tianhao-Gu:
> Why have extensions at all then?

To keep them different from `.bashrc` and `.bash_profile`. It's just a file extension that only makes sense to us and NERSC once had it too so I borrowed it.

MrCreosote:
If you're using them as a template and making the actual files in the user home dir why do the have to have different file names?

Tianhao-Gu:
> If you're using them as a template and making the actual files in the user home dir why do the have to have different file names?

They don't have to be. They're just placeholders. Can you suggest you preferred file names for this PR approval? And I'll just update them accordingly.

MrCreosote:
I would either do no extention at all, or if you expect them to be jinja templates or something, `.tmpl`

Tianhao-Gu:
> I would either do no extension at all, or if you expect them to be jinja templates or something, `.tmpl`

👍

MrCreosote (approved):

PR #108, opened 2024-10-16
create bashrc file

Tianhao-Gu:


src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -49,6 +51,29 @@ def _ensure_user_directory(self, user_dir: Path, username: str):
...
+
+        shutil.copy2(bashrc_ext_src, user_dir/'.bashrc.tmpl')
+        shutil.copy2(bash_profile_ext_src, user_dir/'.bash_profile.tmpl')
+
+        bashrc_dest = user_dir / '.bashrc'
+        bash_profile_dest = user_dir / '.bash_profile'
+
+        if not bashrc_dest.exists():
+            self.log.info(f'Creating .bashrc file for {user_dir}')
+            shutil.copy2(bashrc_ext_src, bashrc_dest)
```
MrCreosote, line 71: Why copy twice? Seems like just copying from the config dir directly to *_dest would work

MrCreosote: Can you add a comment to that effect? Will probably confuse future ppl working on this and they might simplify things w/o realizing your intent

MrCreosote (approved):

PR #109, opened 2024-10-18
update Jupyterhub user guide

Tianhao-Gu:


MrCreosote (approved):
It'd be nice to have instructions on how to create a new venv and switch to it, but that can be another pr

PR #111, opened 2024-10-21
update pipfile

Tianhao-Gu:


Pipfile
```diff
@@ -16,32 +17,28 @@ pyspark= "==3.5.3"
...
 biopython = "==1.84"
 seaborn = "==0.13.2"
-geopandas = "==1.0.1"
-contextily = "==1.6.2"
-dill = "==0.3.9"
-folium = "==0.17.0"
-geopy = "==2.4.1"
-pygeohydro = "==0.18.0"
-pynhd = "==0.18.0"
-numpy = "==2.1.2"
```
Tianhao-Gu, line 44: Users can now install libraries within their own Python virtual environments, there is no need to install those for every users.

Pipfile
```diff
@@ -16,32 +17,28 @@ pyspark= "==3.5.3"
...
 urllib3 = "==2.2.3"
 jupyterlab-git = "==0.50.1"
 cdm-jupyterlab-brand-extension = {file = "https://github.com/kbaseincubator/cdm_jupyterlab_brand_extension/raw/main/dist/cdm_jupyterlab_brand_extension-0.1.0-py3-none-any.whl"}
 
+# Jupyter AI libraries
+jupyter-ai = "==2.25.0"
+langchain-openai = "==0.1.25"
+huggingface_hub = "==0.26.0"
+
 # Frequently utilized scientific libraries by the science team
```
MrCreosote, line 30: Is this comment still relevant?

MrCreosote (approved):

Pipfile
Tianhao-Gu: 👍 updated comment.

PR #112, opened 2024-10-22
enable jupyter ai

Tianhao-Gu:


MrCreosote (approved):
I'd suggest making the file name an env var since it's used in several places

Tianhao-Gu:
> I'd suggest making the file name an env var since it's used in several places

👍

PR #116, opened 2024-11-01
update minio IP

Tianhao-Gu:


MrCreosote (approved):
Do we know why the name doesn't work any more? The new minio is on ci04 I think

Tianhao-Gu:
> Do we know why the name doesn't work any more? The new minio is on ci04 I think

not exactly sure. But that's also the reason many things failed (yarn, query, etc) when we switched.

PR #117, opened 2024-11-12
try not limit container resource

Tianhao-Gu:


config/jupyterhub_config.py
```diff
@@ -50,6 +50,10 @@
 c.DockerSpawner.cpu_limit = 4
 c.DockerSpawner.mem_limit = '16G'
 
+c.DockerSpawner.http_timeout = 120 # 2 minutes (default is 30 seconds)
+c.DockerSpawner.start_timeout = 300  # 5 minutes (default is 60 seconds)
```
MrCreosote, line 54: Could you add comments describing what each timeout is waiting for? Is http_timeout waiting to make contact with the new container, and start waiting for the container to publish it's status as started or something?

Tianhao-Gu: added a ref link in the overhead section. I don’t think every setting needs to be documented.

MrCreosote (approved):

PR #118, opened 2024-11-18
add idle timeout

Tianhao-Gu:


src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -41,6 +44,59 @@ def start(self):
...
+        Returns:
+            timedelta: Idle timeout duration.
+        """
+        idle_timeout_minutes = int(os.getenv("IDLE_TIMEOUT_MINUTES", self.DEFAULT_IDLE_TIMEOUT_MINUTES))
+        self.log.info(f"Idle timeout set to {idle_timeout_minutes} minutes")
+        return timedelta(minutes=idle_timeout_minutes)
+
+    async def poll(self):
+        """
+        Overrides the poll method to periodically checks the status of the user’s JupyterHub container.
```
MrCreosote, line 61:
```suggestion
        Overrides the poll method to periodically check the status of the user’s JupyterHub container.
```

src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -41,6 +44,59 @@ def start(self):
...
+                         to inactivity. Returns None if the container is still active
+                         and running.
+        """
+        # Check if the container has already stopped
+        status = await super().poll()
+        if status is not None:
+            # Container has already stopped, return its status code immediately
+            return status
+
+        last_activity = self.user.last_activity
```
MrCreosote, line 85: So this must be running in the central hub server since it's polling against possibly stopped containers. Is there a CustomDockerSpawner instance per user?

src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -41,6 +44,59 @@ def start(self):
...
+        self.log.info(f"Last activity for {self.container_name}: {last_activity}")
+
+        if last_activity:
+            idle_time = datetime.utcnow() - last_activity
```
MrCreosote, line 89: `utcnow()` isn't recommended: https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

MrCreosote: Are you sure timezone info isn't necessary?

src/jupyterhub_config/custom_docker_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_docker_spawner.py
Tianhao-Gu: yea, both `utcnow()` and `now()` are working as expected locally. `last_activity = self.user.last_activity` doesn't have timezone info either.

MrCreosote: :+1:

MrCreosote (approved):

PR #119, opened 2024-11-20
clean docker compose file

Tianhao-Gu:


docker-compose.yaml
```diff
@@ -231,6 +160,7 @@ services:
       - NETWORK_NAME=cdm-jupyterhub-network
       - JUPYTERHUB_USER_IMAGE=cdm-jupyterhub-cdm_jupyterhub:latest
       - JUPYTERHUB_MOUNT_BASE_DIR=/cdr/cdm/jupyter
+      - ENVIRONMENT=dev
```
MrCreosote, line 163: Is there no user hub instance?

docker-compose.yaml
```diff
@@ -57,10 +54,7 @@ services:
   spark-worker-1:
     # The latest image from cdm-jupyterhub that includes spark standalone mode
     image: ghcr.io/kbase/cdm-jupyterhub:pr-74
-    container_name: spark-worker-1
     platform: linux/amd64
-    depends_on:
-      - spark-master
```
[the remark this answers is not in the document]
Tianhao-Gu: It really doesn't matter now. I think at one time we might even remove stand alone spark. It slows down docker compose up/down. I can add it back but I think it's okay to leave it out.

docker-compose.yaml
Tianhao-Gu: no. Just one hub instance and permissions are controlled via user groups.

MrCreosote: What does the ENV variable do in that case?

Tianhao-Gu: https://github.com/kbase/cdm-jupyterhub/blob/main/config/jupyterhub_config.py#L64-L66

MrCreosote (approved):

docker-compose.yaml
MrCreosote: I'd probably give that env var a more descriptive name in that case, like REMOVE_CONTAINER or something

PR #121, opened 2025-01-08
remove spark timeout setting

Tianhao-Gu:
Since JupyterHub now automatically detects container activity and terminates idle containers, this setting is no longer necessary.


MrCreosote (commented):
LGTM

MrCreosote (approved):
oops

PR #122, opened 2025-02-03
add error checking for KBase auth

Tianhao-Gu:


MrCreosote (approved):
This looks fine, but might be overkill for what you're doing depending on what you're planning. The error system for collections integrated with the fastapi error handlers to return error codes with any error data structures, which you may not need here. Also, I've abandoned this approach for CTS and instead of embedding error codes in the exceptions I map exceptions to error codes in the fastapi error handlers.

Tianhao-Gu:
> This looks fine, but might be overkill for what you're doing depending on what you're planning. The error system for collections integrated with the fastapi error handlers to return error codes with any error data structures, which you may not need here. Also, I've abandoned this approach for CTS and instead of embedding error codes in the exceptions I map exceptions to error codes in the fastapi error handlers.

👍 Sounds good. I am pretty comfortable using this approach. Maybe overkill but not a huge deal to me.

PR #123, opened 2025-02-03
add kb_user model

Tianhao-Gu:


src/service/kb_user.py
MrCreosote, line 1: I ditched this for a NamedTuple in CTS

MrCreosote (approved):

PR #126, opened 2025-02-03
Add KBase authentication form to JupyterHub login page

dauglyon:
Add KBase login form template

(Incorrectly) configure template path in JupyterHub 

Remove JUPYTER_MODE env var

Add KBase auth script to login page

### DIFF
--- a/config/jupyterhub_config.py
+++ b/config/jupyterhub_config.py
@@ -22,6 +22,10 @@ c.JupyterHub.authenticator_class = 'native'
 c.JupyterHub.template_paths = [
     os.environ['JUPYTERHUB_TEMPLATES_DIR'],
 ]
+kbase_env = (os.environ['JUPYTERHUB_KB_ENV']).lower() if 'JUPYTERHUB_KB_ENV' in os.environ else "ci"
+c.JupyterHub.template_vars = {
+    'kbase_origin': 'https://narrative.kbase.us' if kbase_env == "prod" else f'https://{kbase_env}.kbase.us'
+}
 # ref: https://native-authenticator.readthedocs.io/en/latest/options.html
 c.NativeAuthenticator.open_signup = True
 c.NativeAuthenticator.check_common_password = True
--- a/config/jupyterhub_config.py
+++ b/config/jupyterhub_config.py
@@ -20,7 +20,7 @@ c = get_config()
 # ref: https://native-authenticator.readthedocs.io/en/latest/quickstart.html
 c.JupyterHub.authenticator_class = 'native'
 c.JupyterHub.template_paths = [
-    os.path.abspath("/templates"),
+    os.environ['JUPYTERHUB_TEMPLATES_DIR'],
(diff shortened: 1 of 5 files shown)


docker-compose.yaml
```diff
@@ -171,6 +170,7 @@ services:
       - ./cdr/cdm/jupyter/jupyterhub_secrets:/jupyterhub_secrets
       - ./cdr/cdm/jupyter/jupyterhub/users_home:/jupyterhub/users_home
       - /var/run/docker.sock:/var/run/docker.sock
+      - ./templates:/templates
```
Tianhao-Gu, line 173:
```suggestion
```

config/jupyterhub_config.py
```diff
@@ -19,7 +19,10 @@
 # Set the authenticator class to nativeauthenticator
 # ref: https://native-authenticator.readthedocs.io/en/latest/quickstart.html
 c.JupyterHub.authenticator_class = 'native'
-c.JupyterHub.template_paths = [f"{os.path.dirname(nativeauthenticator.__file__)}/templates/"]
+c.JupyterHub.template_paths = [
+    os.path.abspath("/templates"),
```
Tianhao-Gu, line 23:
```suggestion
    os.environ['JUPYTERHUB_TEMPLATES_DIR'],
```

Tianhao-Gu: in Dockerfile, add below before `# Jupyter Hub user home directory`

```# Set up JupyterHub login templates directory
ENV JUPYTERHUB_TEMPLATES_DIR=/templates
RUN mkdir -p ${JUPYTERHUB_TEMPLATES_DIR}
COPY ./templates/ ${JUPYTERHUB_TEMPLATES_DIR}
```

Tianhao-Gu (commented):
We just need to copy file into container via Dockerfile instead of setting it via docker-compose file.

config/jupyterhub_config.py
dauglyon: Sweet!!

templates/login.html
```diff
@@ -0,0 +1,134 @@
...
+    }
+    // setup onSubmit feedback
+    $('form').submit((e) => {
+      var form = $(e.target);
+      form.find('.feedback-container>input').attr('disabled', true);
+      form.find('.feedback-container>*').toggleClass('hidden');
+      form.find('.feedback-widget>*').toggleClass('fa-pulse');
+    });
+
+    const enviromentOrigin = "https://ci.kbase.us"
```
Tianhao-Gu, line 89: Can we make it dynamic? I can pass a env var something like `KB_ENV`, it will be either CI, APPDEV or PROD.

dauglyon: Yup! This was just a placeholder

dauglyon: Need to lookup how to get env variables into jinja and how to special case production, but I think it'll be straightforward

Tianhao-Gu (approved):
nice work!

PR #127, opened 2025-02-04
add custom Jupyterhub KBase auth

Tianhao-Gu:


src/jupyterhub_config/kb_jupyterhub_auth.py
```diff
@@ -0,0 +1,77 @@
...
+        help="KBase Auth2 API URL (e.g., https://ci.kbase.us/services/auth/)"
+    )
+
+    auth_full_admin_roles = List(
+        default_value=[
+            role.strip() for role in os.getenv("AUTH_FULL_ADMIN_ROLES", "").split(",") if role.strip()
+        ],
+        config=True,
+        help="Comma-separated list of KBase roles with full administrative access to JupyterHub."
+    )
```
MrCreosote, line 44: I don't really understand this setup, can you explain?

src/jupyterhub_config/kb_jupyterhub_auth.py
```diff
@@ -0,0 +1,77 @@
...
+    async def authenticate(self, handler, data=None) -> dict:
+        """
+        Authenticate user using KBase session cookie and API validation
+        """
+        session_token = handler.get_cookie(self.SESSION_COOKIE_NAME)
+
+        if not session_token:
+            raise MissingTokenError(f"Authentication required - missing {self.SESSION_COOKIE_NAME} cookie.")
+
+        kb_auth = await KBaseAuth.create(self.kbase_auth_url, self.auth_full_admin_roles)
```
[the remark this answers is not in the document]
Tianhao-Gu: it only created once when users login.

src/jupyterhub_config/kb_jupyterhub_auth.py
Tianhao-Gu: With this setting, you can overwrite the setting in `jupyterhub_config` by doing something like 
`c.JupyterHub.authenticator_class. auth_full_admin_roles`.

src/jupyterhub_config/kb_jupyterhub_auth.py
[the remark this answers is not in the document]
MrCreosote: This is in the hub right? So it's created for every login

src/jupyterhub_config/kb_jupyterhub_auth.py
MrCreosote: So the traitlets + config=True tells jupyterhub to add these variables to the config, am I understand that correctly?

src/jupyterhub_config/kb_jupyterhub_auth.py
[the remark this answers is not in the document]
Tianhao-Gu: I think I see what you mean now. I think it's not worth the trouble here as I need to manage existing event loop if I create kb_auth_client outside this class.

src/jupyterhub_config/kb_jupyterhub_auth.py
Tianhao-Gu: Yea. And config=True allows you to reset it via config.py. Still learning but it's documented on their website via https://jupyterhub.readthedocs.io/en/latest/reference/authenticators.html#authenticator-authenticate

src/jupyterhub_config/kb_jupyterhub_auth.py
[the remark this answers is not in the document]
MrCreosote: Can you at least disable the url check on creation in the client in that case? It's totally pointless if you're not saving the the client for reuse. 

Also caching is pointless, but at least that's not making an extraneous call to the auth server

src/jupyterhub_config/kb_jupyterhub_auth.py
MrCreosote: :+1:

MrCreosote: You might want to add that link as a comment to the class btw

src/jupyterhub_config/kb_jupyterhub_auth.py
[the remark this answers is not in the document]
Tianhao-Gu: Added a TODO item for future refactoring, as it falls outside the current concerns of the project.

src/jupyterhub_config/kb_jupyterhub_auth.py
Tianhao-Gu: 👍

src/jupyterhub_config/kb_jupyterhub_auth.py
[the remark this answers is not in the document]
Tianhao-Gu: 👍 fine. I think it's really not our concern at the moment, nor does it significantly reduce the load on the Auth2 server, regardless of how easy or difficult the proposed change to block PRs may be.

MrCreosote (approved):

PR #128, opened 2025-02-05
Simplify auth config, set c.JupyterHub.authenticator_class

dauglyon:
# Before
Screenshot 2025-02-04 at 5 13 39 PM
# After
Screenshot 2025-02-04 at 5 14 25 PM

### DIFF
--- a/config/jupyterhub_config.py
+++ b/config/jupyterhub_config.py
@@ -10,21 +10,18 @@ https://jupyterhub-dockerspawner.readthedocs.io/en/latest/api/index.html#dockers
 
 import os
 
-import nativeauthenticator
-
 from jupyterhub_config.custom_docker_spawner import CustomDockerSpawner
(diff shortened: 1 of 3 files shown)


Tianhao-Gu (approved):
LGTM. Do you think we could cutover auth completely? I have something like `if os.environ.get('USE_KBASE_AUTHENTICATOR', 'false').lower() == 'true':` locally in case we need to rollback. But I am completely fine with a complete cutover if you think the front end already does the trick.

dauglyon:
I don't think we need to maintain the username/pswd login if this one is working, as long as the users map over so people still have access to their data. Is there work needed there?

Tianhao-Gu:
> I don't think we need to maintain the username/pswd login if this one is working, as long as the users map over so people still have access to their data. Is there work needed there?

I think we should be fine. I've been asking people to register with their KBase user name. (Most people did that). There maybe some manual one-off tasks.

PR #130, opened 2025-02-06
Add update loop and login nextRequest return leg param

dauglyon:


templates/login.html
```diff
@@ -86,21 +86,38 @@ <h1>Sign in</h1>
...
-        const buttonText = 'Log in with KBase as "' + username + '"';
-        $('#login_submit').attr('value', buttonText);
-        $('#login_submit').toggleClass('hidden', false);
-        $('#login_not_ok').toggleClass('hidden', true);
-      } else {
-        $('#login_submit').toggleClass('hidden', true);
-        $('#login_not_ok').toggleClass('hidden', false);
+    // Initial state update
+    updateState();
+    // Check for cookie changes in other windows w/ useInterval,
```
dauglyon, line 94:
```suggestion
    // Check for cookie changes in other windows w/ setInterval,
```

Tianhao-Gu (approved):
I am pretty useless in Javascript. But I can confirm I was able to get the redirect page working locally, so I believe it should be good to go.

PR #131, opened 2025-02-06
(Optional) Typescript

dauglyon:


.gitignore
```diff
@@ -12,3 +12,7 @@ build
 /.pydevproject*
 /.pytest_cache/
 /.settings/
+
+# Ignore ts build folders
+/ts/node_modules
+/ts/build
```
dauglyon, line 18:
```suggestion
/ts/ts-build
```

Tianhao-Gu (approved):
Verified locally that login workflow works as expected.

PR #134, opened 2025-02-24
CDM-82 add kubspawner

Tianhao-Gu:
Primarily copied from `custom_docker_spawner.py`, except for differences in how Kubernetes mounts volumes (`_ensure_user_volume`). I didn’t create a common utility to abstract shared functions since we’ll eventually only have one spawner class. I also don't need to worry about compatibility issues for the Kubernetes spawner development.

### DIFF
--- a/src/jupyterhub_config/custom_kube_spawner.py
+++ b/src/jupyterhub_config/custom_kube_spawner.py
@@ -91,7 +91,7 @@ class CustomKubeSpawner(KubeSpawner):
             if idle_time > self.idle_timeout:
                 self.log.warn(f"Pod {self.pod_name} has been idle for {idle_time}. Stopping...")
                 await self.stop()
-                return 0
+                return 0  # Return an exit code to indicate the container has stopped
 
         # Return status (None) to indicate that the container is still running and active
         return status
--- a/src/jupyterhub_config/custom_kube_spawner.py
+++ b/src/jupyterhub_config/custom_kube_spawner.py
@@ -246,13 +246,14 @@ class CustomKubeSpawner(KubeSpawner):
         mount_base_dir = os.environ['JUPYTERHUB_MOUNT_BASE_DIR']
         hub_secrets_dir = os.environ['JUPYTERHUB_SECRETS_DIR']
 
-        cdm_shared_dir = os.environ['CDM_SHARED_DIR']
-        hive_metastore_dir = os.environ['HIVE_METASTORE_DIR']
-        kbase_shared_dir = os.environ['KBASE_GROUP_SHARED_DIR']
+        cdm_shared_dir = os.environ['CDM_SHARED_DIR']  # Legacy data volume from JupyterLab
+        hive_metastore_dir = os.environ['HIVE_METASTORE_DIR']  # within cdm_shared_dir
+        kbase_shared_dir = os.environ['KBASE_GROUP_SHARED_DIR']  # within cdm_shared_dir
 
         if self.user.admin:
             self.log.info(f"Admin user detected: {self.user.name}. Setting up admin volume mounts.")
             self.volumes = [
+                # Global users home directory
                 {
                     "name": "user-home",
                     "hostPath": {"path": f"{mount_base_dir}/{user_home_dir}"}
@@ -261,6 +262,7 @@ class CustomKubeSpawner(KubeSpawner):
                     "name": "jupyterhub-secrets",
                     "hostPath": {"path": f"{mount_base_dir}/{hub_secrets_dir}"}
                 },
+                # Legacy data volume from JupyterLab
                 {
                     "name": "cdm-shared",
                     "hostPath": {"path": f"{mount_base_dir}/{cdm_shared_dir}"}
@@ -280,10 +282,12 @@ class CustomKubeSpawner(KubeSpawner):
                     "name": "hive-metastore",
                     "hostPath": {"path": f"{mount_base_dir}/{hive_metastore_dir}"}
                 },
+                # User specific home directory
                 {
                     "name": "user-home",
                     "hostPath": {"path": f"{mount_base_dir}/{user_home_dir}/{self.user.name}"}
                 },
+                # Legacy data volume from JupyterLab
                 {
                     "name": "kbase-shared",
                     "hostPath": {"path": f"{mount_base_dir}/{kbase_shared_dir}"}
--- a/src/jupyterhub_config/custom_kube_spawner.py
+++ b/src/jupyterhub_config/custom_kube_spawner.py
@@ -1,0 +1,345 @@
+import os
+import shutil
+import venv
+import json5
+from datetime import timedelta, datetime
+from pathlib import Path
+from filelock import FileLock
+
+from kubespawner import KubeSpawner
+
+class CustomKubeSpawner(KubeSpawner):
+    RW_MINIO_GROUP = 'minio_rw'
+    DEFAULT_IDLE_TIMEOUT_MINUTES = 180
+
+    def start(self):
+        username = self.user.name
+        global_home = Path(os.environ['JUPYTERHUB_USER_HOME'])
+        user_dir = global_home / username
+        self.idle_timeout = self._get_idle_timeout()
+
+        # Ensure the user directory exists
+        self._ensure_user_directory(user_dir, username)
+        self._ensure_bashrc(user_dir)
+
+        # Ensure the user's Jupyter directory exists
+        self._ensure_user_jupyter_directory(user_dir)
+
+        # Ensure the virtual environment is created or reused
+        user_env_dir = user_dir / '.virtualenvs' / 'envs' / f'{username}_default_env'
+        self._ensure_virtual_environment(user_env_dir)
+
+        # Configure environment variables for the pod
+        self._configure_environment(user_dir, user_env_dir, username)
+
+        # Configure notebook directory
+        self._configure_notebook_dir(username, user_dir)
+
+        # Set up volume mounts specific for Kubernetes
+        self._ensure_user_volume(user_dir)
+
+        # Set up JupyterLab favorites (if applicable)
+        self._add_favorite_dir(user_dir, favorites={Path(os.environ['KBASE_GROUP_SHARED_DIR'])})
+
+        self.namespace = os.environ['KUBE_NAMESPACE']
+
+        return super().start()
+
+    def _get_idle_timeout(self):
+        """
+        Retrieves the idle timeout from the environment variable `IDLE_TIMEOUT_MINUTES`.
+        If not set, defaults to 180 minutes.
+
+        Returns:
+            timedelta: Idle timeout duration.
+        """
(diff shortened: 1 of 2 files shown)


src/jupyterhub_config/custom_kube_spawner.py
```diff
@@ -0,0 +1,345 @@
...
+                         and running.
+        """
+        # Check if the container has already stopped
+        status = await super().poll()
+        if status is not None:
+            # Container has already stopped, return its status code immediately
+            return status
+
+        last_activity = self.user.last_activity
+        self.log.info(f"Last activity for {self.pod_name}: {last_activity}")
```
MrCreosote, line 87: is pod_name set by the parent class?

src/jupyterhub_config/custom_kube_spawner.py
```diff
@@ -0,0 +1,345 @@
...
+            try:
+                # Create a virtual environment with system site-packages access
+                venv.create(env_dir=user_env_dir, system_site_packages=True, with_pip=True)
+            except Exception as e:
+                raise ValueError(f'Failed to create virtual environment for {self.user.name}: {e}') from e
+        else:
+            self.log.info(f'Reusing virtual environment for {self.user.name}')
+
+    def _configure_environment(self, user_dir: Path, user_env_dir: Path, username: str):
+        self.environment.update({key: str(value) for key, value in os.environ.items() if key not in self.environment})
```
MrCreosote, line 175: Is it possible to have non str values in os.environ?

src/jupyterhub_config/custom_kube_spawner.py
```diff
@@ -0,0 +1,345 @@
...
+        self.environment['JUPYTER_MODE'] = 'jupyterhub-singleuser'
+        # self.environment['JUPYTERHUB_ADMIN'] = str(self.user.admin)
```
MrCreosote, line 177: I don't totally understand why this is not longer needed

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote, line 1: There are various comments throughout the file that were removed that were, at least in some cases, pretty useful. Can they be put back in?

MrCreosote: Same for a TODO here and there that are probably good ideas

src/jupyterhub_config/custom_kube_spawner.py
```diff
@@ -0,0 +1,345 @@
...
+        group_names = [group.name for group in self.user.groups]
+        return self.user.admin or self.RW_MINIO_GROUP in group_names
+
+    def _ensure_user_volume(self, user_dir: Path):
+        """
+        Ensure the user's volume is correctly mounted in the container.
+        """
+
+        user_home_dir = os.environ['JUPYTERHUB_USER_HOME']
+        mount_base_dir = os.environ.get('JUPYTERHUB_MOUNT_BASE_DIR')
```
MrCreosote, line 235: Under what situations will this not be available?

src/jupyterhub_config/custom_kube_spawner.py
```diff
@@ -0,0 +1,345 @@
...
+        cdm_shared_dir = os.environ['CDM_SHARED_DIR']
+        hive_metastore_dir = os.environ['HIVE_METASTORE_DIR']
+        kbase_shared_dir = os.environ['KBASE_GROUP_SHARED_DIR']
+
+        if self.user.admin:
+            self.log.info(f"Admin user detected: {self.user.name}. Setting up admin volume mounts.")
+            self.volumes = [
+                {
+                    "name": "user-home",
+                    "hostPath": {"path": f"{mount_base_dir}/{user_home_dir}"}
```
MrCreosote, line 247: mount_base_dir might be None here

src/jupyterhub_config/custom_kube_spawner.py
```diff
@@ -0,0 +1,345 @@
...
+                {"name": "cdm-shared", "mountPath": cdm_shared_dir}
+            ]
+        else:
+            self.log.info(f"Non-admin user detected: {self.user.name}. Setting up user-specific volume mounts.")
+            # Determine readOnly mode: True if NOT a read/write minio user
+            read_only = not self._is_rw_minio_user()
+            self.volumes = [
+                {
+                    "name": "hive-metastore",
+                    "hostPath": {"path": f"{mount_base_dir}/{hive_metastore_dir}"}
```
MrCreosote, line 270: mount_base_dir again

MrCreosote (commented):
I made a new branch and copied the kube spawner on top of the docker spawner and just reviewed the changes, not unchanged code.

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: yea.

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: No. Kubenetes doesn't allow non-str env variables.

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: I put it back. `self.user.admin` is a bool and gave me trouble before. I commented it and forget to comment it back. It's current just an indicator. Admin user is set by the custom auth now via a KBase Auth role.

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: 👍

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: it should never be none. Probably a copy error.

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: fixed.

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: What's the purpose of stringifying the value in that case?

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: So it is necessary, or can it be removed completely?

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: Can we revert to square brackets then?

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: Just so that we still keep everything in case we need it.

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: we can keep it for troubleshooting I guess. It's not strictly needed.

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: But if os.environ can't have non-str values why bother with the str()?

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: If it's not necessary I'd probably remove it but I leave it up to you

Tianhao-Gu: I will leave it in case we need it. KubeSpawner is pretty fragile now.

MrCreosote: But if you keep it, did you mean to stringify the bool?

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: So that we keep a copy of things in case we need to use it. I don't really get what you mean actually.

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: Yes. non-string value is not allowed to save to `self.environment` for kubespawner.

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: There's a few other comments in the _ensure_user_volume that would be useful noting that the cdm shared dir is legacy and presumably therefore will be removed at some point

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: The code is calling str() on the values of the os.environ dict. If the values of the os.environ dict are required to be strings already, it's a noop and there's no need for the str() call

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: 👍

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: is it used anywhere else where it's assumed to be a bool? E.g. now `if self.environment['JUPYTERHUB_ADMIN']` will always be True

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: Not in our codebase. Inside of spawner, I think it might just use `self.user.admin`.

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: You can still set os.environ whatever you want. `self.environment` doesn't allow non-string values for KubeSpawner.

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_kube_spawner.py
MrCreosote: :+1:

MrCreosote (approved):

PR #135, opened 2025-02-28
update jupyterhub config file

Tianhao-Gu:


config/jupyterhub_config.py
```diff
@@ -13,12 +13,17 @@
...
-if os.environ.get('USE_KBASE_AUTHENTICATOR', 'false').lower() == 'true':
+def get_bool_env(key, default=False):
+    """Parse str boolean environment variables"""
+    value = os.environ.get(key, str(default)).lower()
+    return value in ('true', '1', 't')
+
+if get_bool_env('USE_KBASE_AUTHENTICATOR'):
     # Set the authenticator class to KBaseAuthenticator
     # ref: https://jupyterhub.readthedocs.io/en/latest/reference/authenticators.html#authenticators
     c.JupyterHub.authenticator_class = KBaseAuthenticator
```
MrCreosote, line 29: So this doesn't need to be removed any more? Seems like we're there so the native auth can be dumped...

config/jupyterhub_config.py
```diff
@@ -54,40 +59,58 @@
...
-# Set extra labels in order to use Rancher's network policies
-# ref: https://rancher.com/docs/rancher/v1.6/en/rancher-services/networking/#containers-created-with-the-docker-cli
-c.DockerSpawner.extra_create_kwargs = {
-    'labels': {
-        'io.rancher.container.network': 'true'
-    }
-}
 
-c.DockerSpawner.shutdown_no_activity_timeout = 60 * 60  # 1 hour
+if get_bool_env('USE_KUBE_SPAWNER'):
```
MrCreosote, line 64: Should kube be the default?

config/jupyterhub_config.py
```diff
@@ -54,40 +59,58 @@
...
+    spawner_class = CustomKubeSpawner
+    spawner_config = c.KubeSpawner
+
+    # User Kubernetes pod configuration
+    c.KubeSpawner.namespace = os.environ['KUBE_NAMESPACE']
+    c.KubeSpawner.hub_connect_url = f"http://{os.environ['SPARK_DRIVER_HOST']}:8081"
+
+    def modify_pod_hook(spawner, pod):
+        pod.spec.service_account_name = "cdm-jupyterhub"
+        pod.metadata.labels.update({"app": "cdm-jupyterhub"})
```
MrCreosote, line 79: What do these lines do?

config/jupyterhub_config.py
```diff
@@ -54,40 +59,58 @@
...
+        pod.spec.dns_policy = "ClusterFirstWithHostNet"
+        return pod
+    c.KubeSpawner.modify_pod_hook = modify_pod_hook
+    c.KubeSpawner.image_pull_policy = 'IfNotPresent'
+
+    # For troubleshooting purposes, keep the user pod in non-prod environment
+    c.KubeSpawner.delete_stopped_pods = environment != 'dev'
```
MrCreosote, line 86: Should this be an env var rather than always on / always off per env?

config/jupyterhub_config.py
```diff
@@ -54,40 +59,58 @@
...
+    c.DockerSpawner.mem_limit = '16G'
+
+    # The network name that Docker containers will use to communicate
+    if network_name := os.environ.get('NETWORK_NAME'):
+        c.DockerSpawner.network_name = network_name
+    c.DockerSpawner.use_internal_ip = True
+    c.DockerSpawner.extra_create_kwargs = {'labels': {'io.rancher.container.network': 'true'}}
+
+    # For troubleshooting purposes, keep the user container in non-prod environment
+    c.DockerSpawner.remove = environment != 'dev'
```
MrCreosote, line 104: Same ? as above

config/jupyterhub_config.py
```diff
@@ -54,40 +59,58 @@
...
-
-# Set extra labels in order to use Rancher's network policies
-# ref: https://rancher.com/docs/rancher/v1.6/en/rancher-services/networking/#containers-created-with-the-docker-cli
-c.DockerSpawner.extra_create_kwargs = {
-    'labels': {
-        'io.rancher.container.network': 'true'
-    }
-}
 
-c.DockerSpawner.shutdown_no_activity_timeout = 60 * 60  # 1 hour
```
MrCreosote, line 90: This line seems to be completely gone, is that what you want?

config/jupyterhub_config.py
Tianhao-Gu: I thought we were. But we also have this ticket: https://kbase-jira.atlassian.net/browse/CDM-32 to keep native auth as fallback.

config/jupyterhub_config.py
Tianhao-Gu: yea. No longer needed. It was controlled by spawner for both docker and kube.

config/jupyterhub_config.py
Tianhao-Gu: No strong preference. It doesn't really matter.

config/jupyterhub_config.py
Tianhao-Gu: add a label to pods so that policies/configs can be batch applied in kubernetes.

config/jupyterhub_config.py
Tianhao-Gu: Also no strong option. If you think an env is better, I will change.

config/jupyterhub_config.py
MrCreosote: The problem with switching between kbase auth and local auth is that if 2 different people have the same name in the different auth systems, the ownership of their files will switch back and forth, IIUC. That seems Really Bad.

config/jupyterhub_config.py
MrCreosote: If it's our intention to switch to kube permanently and remove the docker spawner I'd say make kube the default. If the plan is to keep both, I don't have strong feelings either.

config/jupyterhub_config.py
MrCreosote: could you add a comment to that effect please?

config/jupyterhub_config.py
MrCreosote: I personally would go with env as you might need to debug in a non-dev environment, but you're the one that's probably going to be affected by that, so I leave it up to you

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
Tianhao-Gu: yea. I didn't create the ticket. Maybe we just want to leave a fallback option incase anything went wrong.

config/jupyterhub_config.py
Tianhao-Gu: 👍  we are still in exploration mode, nothing is concrete. I made default to use Kubernetes. I think it doesn't matter if we ensure to set the variable in every deployment.

config/jupyterhub_config.py
Tianhao-Gu: 👍

config/jupyterhub_config.py
Tianhao-Gu: 👍

config/jupyterhub_config.py
Tianhao-Gu: 👍

config/jupyterhub_config.py
MrCreosote: I added a comment on the ticket: https://kbase-jira.atlassian.net/browse/CDM-32?focusedCommentId=98362

would you mind adding that link as a comment in the code? I wouldn't want future people to get the wrong impression that it's ok to switch back and forth

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
MrCreosote: :+1:

config/jupyterhub_config.py
Tianhao-Gu: 👍

MrCreosote (approved):

PR #151, opened 2025-03-06
try to use IP as driver hostname

Tianhao-Gu:


src/spark/utils.py
```diff
@@ -113,6 +114,14 @@ def get_spark_session(
 
     spark_conf = _get_base_spark_conf(app_name, executor_cores, yarn)
     sc = {}
+    if os.environ.get('USE_KUBE_SPAWNER') == 'true':
+        yarn = False  # YARN is not used in the Kubernetes spawner
+        # Since the Spark driver cannot resolve a pod's hostname without a dedicated service for each user pod,
+        # use the pod IP as the identifier for the Spark driver host
+        hostname = os.environ["SPARK_DRIVER_HOST"]
+        if hostname:
```
MrCreosote, line 122: Should this throw an error if there's no host?

src/spark/utils.py
```diff
@@ -113,6 +114,14 @@ def get_spark_session(
 
     spark_conf = _get_base_spark_conf(app_name, executor_cores, yarn)
     sc = {}
+    if os.environ.get('USE_KUBE_SPAWNER') == 'true':
```
MrCreosote, line 117: Hmm, your codebase so up to you, but I would make this an argument rather than pulling from env vars inside a fn

src/spark/utils.py
```diff
@@ -113,6 +114,14 @@ def get_spark_session(
...
+        yarn = False  # YARN is not used in the Kubernetes spawner
+        # Since the Spark driver cannot resolve a pod's hostname without a dedicated service for each user pod,
+        # use the pod IP as the identifier for the Spark driver host
+        hostname = os.environ["SPARK_DRIVER_HOST"]
```
MrCreosote, line 121: Or just pass in the hostname

src/spark/utils.py
Tianhao-Gu: I think that's fine. In kubespawner, this would never happen. `SPARK_DRIVER_HOST` is set in kubespawner.

src/spark/utils.py
Tianhao-Gu: I need to make `get_spark_session` backwards compatible. Users are currently calling `spark = get_spark_session()`- this needs to be working after updates without changes from user's end.

src/spark/utils.py
MrCreosote: In that case, it seems like if there's no host name something has gone Really Wrong and throwing an error would be the right thing to do

MrCreosote (approved):

PR #152, opened 2025-03-12
enable fair scheduler

Tianhao-Gu:


Dockerfile
MrCreosote, line 1: What fair scheduler does this affect?

config/spark-fairscheduler.xml
```diff
@@ -0,0 +1,13 @@
+<allocations>
+    <pool name="default">
+        <schedulingMode>FAIR</schedulingMode>
+        <weight>1</weight>
+        <minShare>1</minShare>
+    </pool>
+
+    <pool name="highPriority">
```
MrCreosote, line 8: Can any user choose the high priority pool?

src/spark/utils.py
```diff
@@ -142,6 +146,11 @@ def get_spark_session(
         spark_conf.set(key, value)
     spark = SparkSession.builder.config(conf=spark_conf).getOrCreate()
 
+    if scheduler_pool not in SPARK_POOLS:
+        print(f"Warning: Scheduler pool {scheduler_pool} is not in the list of available pools: {SPARK_POOLS} "
+              f"Defaulting to 'default' pool")
+    spark.sparkContext.setLocalProperty("spark.scheduler.pool", scheduler_pool)
```
MrCreosote, line 152:
```suggestion
    if scheduler_pool not in SPARK_POOLS:
        print(f"Warning: Scheduler pool {scheduler_pool} is not in the list of available pools: {SPARK_POOLS} "
              f"Defaulting to 'default' pool")
        scheduler_pool = "default"  # might want to make this a constant
    spark.sparkContext.setLocalProperty("spark.scheduler.pool", scheduler_pool)
```

config/spark-fairscheduler.xml
Tianhao-Gu: yea. Currently yes. Maybe we can configure some user grouping in the future. But that's up to a discussion.

Dockerfile
Tianhao-Gu: hmm I didn't get your question. Here is the link for fair scheduler pool if you are interested:
 https://spark.apache.org/docs/latest/job-scheduling.html#fair-scheduler-pools

src/spark/utils.py
MrCreosote: So if someone passes in "my_special_pool" then spark will just ignore it and use "default"? It's not clear that's what happens from the code

config/spark-fairscheduler.xml
MrCreosote: Ok, as long as Gazi is ok with it, I'm ok with it

src/spark/utils.py
Tianhao-Gu: 👍 I don't see it's necessary personally but sure...

Dockerfile
MrCreosote: What I don't understand here is that this looks like it's all configured client side, so how would fair share be enforced in standalone / yarn / k8s? 

It looks to me that this fair scheduler config *only* works within a single spark context. If that's correct, given every user gets their own context, I'm not sure this does what you want it to do

https://spark.apache.org/docs/latest/job-scheduling.html#scheduling-within-an-application

config/spark-fairscheduler.xml
Tianhao-Gu: I think it's unrelated to this PR. Users can request all available resources currently. We will for sure implementing something that allow/deny users to request high priority queues. But not now.

MrCreosote: Same response as above

src/spark/utils.py
MrCreosote: Whether it's necessary or not depends on what the scheduler does when it gets an invalid pool name

Dockerfile
Tianhao-Gu: What's your suggestion then? If we don't do this.

src/spark/utils.py
Tianhao-Gu: yea. Like I already said `the spark context automatically assign the pool to default if scheduler_pool is unavailable.`

Dockerfile
MrCreosote: I don't know, I don't have any knowledge of this stuff other than the YARN fair scheduler, which is already set up

src/spark/utils.py
MrCreosote: What does unavailable mean though? Null or some unrecognized string?

Dockerfile
Tianhao-Gu: This is similar to YARN fair scheduler I think. There are different pools and users have equal shares of resources. Part of the reasons we are doing this is because no one understand/support YARN.

src/spark/utils.py
Tianhao-Gu: I updated the code to explicitly assign `scheduler_pool`.

Dockerfile
MrCreosote: I don't think so - like I say above, this seems to only do fair share within a single spark context, if I'm reading the docs correctly.

src/spark/utils.py
MrCreosote: I know, but you're saying that's unnecessary. I'm trying to understand why you're saying that, and if that's correct, I agree we should remove it (although leave a comment to prevent the next person having the same question as I do)

Dockerfile
Tianhao-Gu: It's on the client's side. I don't get your point/concern why it matters whether it's within or not within a single spark context. But this enabled us to have different pools so that jobs with different purpose can have different resources. Users can request all cores with YARN even though YARN's fair scheduler is configured on the server side.

Tianhao-Gu: The overall resources are managed by Kubernetes. In K8S, we limit each workers' resources. It has its own native scheduling to ensure a fair share of resources. But this is also unrelated to this pooling PR. 

```
        resources:
          requests:
            memory: "1.5Gi"
            cpu: "2.5"
          limits:
            memory: "2Gi"
            cpu: "3"
```

MrCreosote: > I don't get your point/concern why it matters whether it's within or not within a single spark context

You want fair share set up so that if user A is hogging resources, then user B's jobs will start to take priority, right? If the fair share only works in a single spark context, and each user has their own spark context, then fair share won't do anything about user B hogging resources.

Tianhao-Gu: yes you are right but I don't see it has anything to do with this PR. Cross users application resources are managed by K8S. This update is to enable a default pool with 1 core per job by default and having ability for users/jobs to bump up resources.

MrCreosote: Ok, the title "enable fair scheduler" made me think the purpose of this PR was to get a fair scheduler working, which it seems is not the case. 

Given that, I don't really understand what this PR does. You create a default and high priority pool within the context, which presumably the actual scheduler doesn't know about (?), and they do... what exactly?

Tianhao-Gu: Our plan is to maximum spark worker's core count and having clients to use different pools so that they don't request all cores by using the `get_spark_session` method. In other words, if there are 128 cores on a worker, we can support 128 users if they all use the default settings of `get_spark_session`. Without this, the first user will reserve all cores.

Tianhao-Gu: We don't really have an `actual scheduler` in K8S. The workers are sitting there (with maximum capacity) and awaiting for jobs basically.

MrCreosote: Ok, so IIUC, what this is doing is by default giving the users 1 core per job and if they ask for high priority they get 2 cores?

If that's correct, there's nothing stopping them from making their own spark context and running any number of cores they want, right?

Tianhao-Gu: yea. That's correct, high priority pool will get more cores (up to 5). It won't stop people from making their own spark session and request all cores. With YARN today, I noticed I can also request as many cores as I want. We've set `spark.dynamicAllocation.maxExecutors` in both client and worker nodes. However, that doesn't seem to work. That's something I need to figure out.

MrCreosote: > (up to 5)

how does that work? I'm not seeing a setting for that in the PR

MrCreosote: Oh is it weight?

MrCreosote: A link to the description of the xml format somewhere would be helpful

Tianhao-Gu: 👍 added a link for fairscheduler.xml template. 

Oh yea, I mean 5X not exactly 5 cores. But I think in our case, minimum executor is set to 1 and 5X also means 5 cores.

MrCreosote: I was hoping more for a description of what the fields mean rather than a template, which doesn't help me understand what's going on here

Tianhao-Gu: 👍

MrCreosote: Based on the discussion in the standup, we decided to go ahead with the PR since the understanding is that it lowers the minimum cores requested by a spark context from 8 to (3?). It's not completely clear whether it has any effect on the maximum number of cores a context can use - people are reading the documents differently

src/spark/utils.py
MrCreosote: I'm assuming here you don't care whether we leave the reassignment in or not

Tianhao-Gu: 👍 yea. we can leave the assignment there.

PR #154, opened 2025-03-16
reduce img size

Tianhao-Gu:


Dockerfile
```diff
@@ -40,21 +40,20 @@ ENV POSTGRES_JDBC_VER=42.2.23
...
-
-# install python dependencies
+# Install pipenv and Python dependencies with cache cleanup
+RUN pip3 install --no-cache-dir pipenv
 COPY Pipfile* ./
-RUN pipenv sync --system
+RUN pipenv sync --system && pip cache purge
 
-RUN chown -R spark_user:spark /opt/bitnami
+# RUN chown -R spark_user:spark /opt/bitnami
```
Tianhao-Gu, line 56: It creates a 3.6GB layer. I think this was previously required when the image was used as both Spark master and worker. It may no longer be necessary, but I've kept it commented out as a fallback in case any issues arise.

MrCreosote: Could you put in a comment so the next person knows why this was commented out and when it might need to be restored?

Dockerfile
```diff
@@ -11,7 +11,7 @@ USER root
 # https://github.com/bitnami/containers/pull/52661
 RUN groupadd -r spark && useradd -r -g spark spark_user
 
-RUN apt-get update && apt-get install -y \
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
MrCreosote, line 14: Are you sure this has no negative effects? I could imagine a package having a recommended package that isn't required but speeds up certain operations, for example. Some python packages work that way

I'd be surprised if the recommended packages took a significant amount of space, is that the case?

Dockerfile
MrCreosote, line 1: Out of curiosity how much space do these changes save?

Dockerfile
```diff
@@ -40,21 +40,20 @@ ENV POSTGRES_JDBC_VER=42.2.23
...
 
-# install pipenv
-RUN pip3 install pipenv
-
-# install python dependencies
+# Install pipenv and Python dependencies with cache cleanup
+RUN pip3 install --no-cache-dir pipenv
 COPY Pipfile* ./
-RUN pipenv sync --system
+RUN pipenv sync --system && pip cache purge
```
MrCreosote, line 54: Should this be `pipenv --clear` rather than pip?

Dockerfile
Tianhao-Gu: 👍

Dockerfile
Tianhao-Gu: Sounds good. Reverted this change. It doesn't save us a lot of spaces either.

Dockerfile
MrCreosote: :+1:

Dockerfile
Tianhao-Gu: From 17+GB to 8.4GB which is still pretty large. The base image is ~2GB itself. Since we are no longer using this image for spark clustering, we could consider using jupyterhub image as base image and install spark which should significantly reduce image size. But since Trivy can run for now, I think it should be fine for the moment.

Dockerfile

Dockerfile

Tianhao-Gu: yup. 👍

MrCreosote (approved):

PR #155, opened 2025-03-18
add supporting funcs for the Tree extension

Tianhao-Gu:


Tianhao-Gu:
> The code as given seems fine, but creating a spark session for every invocation seems really expensive just to retrieve table info from postgres. I'm ignorant here but it seems like there ought to be a better way

yea. I recall there were discussions about a always-running session for jobs. If it's getting too bad, I can also grab table info from the postgres directly.

PR #156, opened 2025-03-19
retrieve db info from postgres directly

Tianhao-Gu:


Pipfile
```diff
@@ -23,6 +23,7 @@ urllib3 = "==2.3.0"
 jupyterlab-git = "==0.51.0"
 cdm-jupyterlab-brand-extension = {file = "https://github.com/kbaseincubator/cdm_jupyterlab_brand_extension/raw/main/dist/cdm_jupyterlab_brand_extension-0.1.0-py3-none-any.whl"}
 cacheout = "==0.16.0"
+psycopg2 = "==2.9.10"
```
MrCreosote, line 26: Seems like you should be using 3: https://pypi.org/project/psycopg/

src/postgres/connection.py
```diff
@@ -0,0 +1,45 @@
...
+    default_host, default_port = postgres_url.split(':')
+
+    # Get connection parameters
+    db_params = {
+        'dbname': dbname or os.environ.get('POSTGRES_DB'),
+        'user': user or os.environ.get('POSTGRES_USER'),
+        'password': password or os.environ.get('POSTGRES_PASSWORD'),
+        'host': host or default_host,
+        'port': port or default_port,
+        'cursor_factory': RealDictCursor
```
MrCreosote, line 39: I would do more error checking on the inputs but I leave that up to you

src/postgres/hive_metastore.py
```diff
@@ -0,0 +1,54 @@
...
+from postgres.connection import get_postgres_connection
+
+
+def get_databases() -> List[str]:
+    """Get list of databases from Hive metastore.
+    
+    Returns:
+        List of database names
+    """
+    with get_postgres_connection() as conn:
```
MrCreosote, line 16: Is it expensive to establish a connection? For instance Mongo advises creating a single client for an entire application and keeping it for the life of the application

src/postgres/hive_metastore.py
```diff
@@ -0,0 +1,54 @@
...
+        with conn.cursor() as cur:
+            cur.execute('''
+                SELECT "TBL_NAME" 
+                FROM "TBLS" t 
+                JOIN "DBS" d ON t."DB_ID" = d."DB_ID" 
+                WHERE d."NAME" = %s
+            ''', (database,))
+            rows = cast(List[RealDictRow], cur.fetchall())
+            return [str(row.get('TBL_NAME')) for row in rows]
```
MrCreosote, line 41: For the case where you want all the db names and their tables might be more efficient to do it in one query vs 1 per db

Pipfile
Tianhao-Gu: 👍

src/postgres/connection.py
Tianhao-Gu: 👍

src/postgres/hive_metastore.py
Tianhao-Gu: It's not expensive and pretty fast. But I am also considering to build a connection pool for the postgres connection. For just retrieving db structure, I think it's fine now since the Tree extension likely to jsut call the method once. But for users executing queries, each spark session opens a postgres connection - which might potentially be a problem. But that's from the spark side.

src/postgres/hive_metastore.py
Tianhao-Gu: I was considering that too. But since those 2 queries to get db name and tables are pretty straight forward and fast - I think it should be fine now.

src/postgres/hive_metastore.py
MrCreosote: if each context keeps a connection open for the life of the context that seems fine. I'd be more concerned if connections are being opened several times a minute

Tianhao-Gu: yea. I think for the this specific use case - Tree extension retrieves db info and display the db structure - we should be fine. I am not expecting the extension would call postgres every frequently. In general, that's something I was worrying too. But our Jupyterhub is not like the traditional API app where you can keep a pool of connections at the beginning and all sessions are configured to use the pool. Each user will have their own container and spark also has its own container. We haven't experienced postgres connection problems yet nor I have figured out a good way to share DB connections across different containers and ensure spark worker wouldn't open new connections. But that being said, I created a [story](https://kbase-jira.atlassian.net/browse/CDM-104) to track this if this becomes a bigger concern.

src/postgres/connection.py
```diff
@@ -0,0 +1,84 @@
...
+        psycopg.Error: If connection to the database fails
+    """
+    postgres_url = os.environ.get('POSTGRES_URL', '')
+    if ':' not in postgres_url:
+        raise ValueError("POSTGRES_URL must be in the format 'host:port'")
+    default_host, default_port = postgres_url.split(':')
+
+    # Get and validate connection parameters
+    final_dbname = dbname or os.environ.get('POSTGRES_DB')
+    _validate_not_empty(final_dbname, "Database name", "POSTGRES_DB")
```
MrCreosote, line 56: you could add line 1 into the function for line 2

src/postgres/connection.py
```diff
@@ -0,0 +1,84 @@
...
+
+    final_user = user or os.environ.get('POSTGRES_USER')
+    _validate_not_empty(final_user, "Database user", "POSTGRES_USER")
+
+    final_password = password or os.environ.get('POSTGRES_PASSWORD')
+    _validate_not_empty(final_password, "Database password", "POSTGRES_PASSWORD")
+
+    final_host = host or default_host
+    final_port = port or default_port
```
MrCreosote, line 65: These could be None or the empty string IIUC

MrCreosote (approved):

src/postgres/connection.py
Tianhao-Gu: 👍 added them to `_validate_not_empty`

src/postgres/connection.py
Tianhao-Gu: can you elaborate more? 
Are you suggesting:
```
    final_dbname = _validate_not_empty(dbname or os.environ.get('POSTGRES_DB'), "Database name", "POSTGRES_DB")
```

MrCreosote:
```
final_var = _validate_not_empty(var, env_var_name, variable_name)
```

so it moves line 1 into the function and the environ is checked in there

src/postgres/connection.py
```diff
@@ -0,0 +1,86 @@
...
+    Returns:
+        A PostgreSQL connection with dict_row row factory
+
+    Raises:
+        ValueError: If POSTGRES_URL environment variable is not properly formatted
+        psycopg.Error: If connection to the database fails
+    """
+    postgres_url = os.environ.get('POSTGRES_URL', '')
+    if ':' not in postgres_url:
+        raise ValueError("POSTGRES_URL must be in the format 'host:port'")
```
MrCreosote, line 51: Actually hold on - this will always error out if the url env is unset. Shouldn't it be ok if host and port are set?

src/postgres/connection.py
Tianhao-Gu: 👍

PR #157, opened 2025-03-25
create module to manager user spark clusters in K8S

Tianhao-Gu:


src/spark/kube_spark_manager.py
```diff
@@ -0,0 +1,161 @@
...
+        "POSTGRES_DB": "PostgreSQL database name",
+        "POSTGRES_URL": "PostgreSQL connection URL"
+    }
+
+    # Default configuration values for cluster settings
+    DEFAULT_WORKER_COUNT = 2
+    DEFAULT_WORKER_CORES = 10
+    DEFAULT_WORKER_MEMORY = "10G"
+    DEFAULT_MASTER_CORES = 10
+    DEFAULT_MASTER_MEMORY = "10G"
```
MrCreosote, line 36: I would assume the master doesn't need a lot of resources, although I'm not educated here

src/spark/kube_spark_manager.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Returns:
+            Dict[str, str]: Dictionary of validated environment variables
+        """
+        missing_vars = []
+        env_values = {}
+
+        for var, description in cls.REQUIRED_ENV_VARS.items():
+            value = os.environ.get(var)
+            if not value:
```
MrCreosote, line 61:
```suggestion
            if not value or not value.strip():
```

src/spark/kube_spark_manager.py
```diff
@@ -0,0 +1,161 @@
...
+                "Missing required environment variables:\n" +
+                "\n".join(f"- {var}" for var in missing_vars)
+            )
+
+        return env_values
+
+    def __init__(self,
+                 username: str,
+                 namespace: Optional[str] = None,
+                 image: Optional[str] = None,
```
MrCreosote, line 76: What's the use case for being able to override the env? Users are never going to call this constructor I assume

src/spark/kube_spark_manager.py
```diff
@@ -0,0 +1,161 @@
...
+                 image_pull_policy: str = DEFAULT_IMAGE_PULL_POLICY):
+        """
+        Initialize the KubeSparkManager with user-specific configuration.
+
+        Args:
```
MrCreosote, line 81:
```suggestion
        Initialize the KubeSparkManager with user-specific configuration.
        This can only be run inside a kubernetes container.

        Args:
```

I think that's true based on reading about `load_incluster_config`

MrCreosote (approved):

src/spark/kube_spark_manager.py
Tianhao-Gu: My design is users will be able to call this and create/scale their own spark clustering based on their group permissions. Maybe not directly calling this constructor but a wrapper function. So they might use this. Another design in my mind is ETL/Jobs might use different images other than the user default spark image.

src/spark/kube_spark_manager.py
Tianhao-Gu: yea. This is just the default settings - I am planning to implement overwrite functions so that we can set those via os.environ.

PR #159, opened 2025-03-25
remove service account from users

Tianhao-Gu:


MrCreosote (approved):
I don't understand the implications of this change at all, so I'm trusting you and Boris here.

PR #175, opened 2025-04-23
warn users before creating cluster

Tianhao-Gu:


src/notebook_utils/startup.py
```diff
@@ -25,3 +25,9 @@
     spark_to_pandas,
     display_namespace_viewer,
 )
+from spark.cluster import (
+    check_api_health,
+    create_cluster,
+    delete_cluster,
+    get_cluster_status,
```
MrCreosote, line 32: IIUC users will be calling these functions?

src/spark/cluster.py
```diff
@@ -119,8 +120,18 @@ def create_cluster(
...
-    # TODO - inform user that a new cluster will overwrite the existing cluster
+
+    print("WARNING: Creating a new Spark cluster will terminate your existing cluster.")
+    print("All active Spark sessions and computations will be lost.")
+    
+    if not force:
+        confirmation = input("Do you want to proceed? [Y/n]: ").strip().lower() or 'y'
+        if confirmation not in ('y', 'yes'):
+            print("Cluster creation aborted.")
+            return None
```
MrCreosote, line 133: This implies a cluster always exists, otherwise the existence of the cluster should be checked and the function allowed to proceed if there isn't one, regardless of force. Is that right?

src/spark/cluster.py
```diff
@@ -119,8 +120,18 @@ def create_cluster(
...
         master_memory: Memory for master (e.g., "10GiB")
+        force: Skip confirmation prompt if True
     """
-    # TODO - inform user that a new cluster will overwrite the existing cluster
+
+    print("WARNING: Creating a new Spark cluster will terminate your existing cluster.")
+    print("All active Spark sessions and computations will be lost.")
+    
+    if not force:
+        confirmation = input("Do you want to proceed? [Y/n]: ").strip().lower() or 'y'
```
MrCreosote, line 130: Killing ongoing computations by default seems dangerous

src/spark/cluster.py
```diff
@@ -119,8 +120,18 @@ def create_cluster(
         worker_memory: Memory per worker (e.g., "10GiB")
         master_cores: CPU cores for master
         master_memory: Memory for master (e.g., "10GiB")
+        force: Skip confirmation prompt if True
     """
-    # TODO - inform user that a new cluster will overwrite the existing cluster
+
+    print("WARNING: Creating a new Spark cluster will terminate your existing cluster.")
+    print("All active Spark sessions and computations will be lost.")
```
MrCreosote, line 127: This will always print on notebook startup IIUC. Is that what you want?

src/notebook_utils/startup.py
Tianhao-Gu: not likely. Only power users should call `create_cluster`. But I think it would be helpful to provide users the ability to call all of them.

src/spark/cluster.py
Tianhao-Gu: added a TODO item to check existence of user's clustering.  I think it's not very important at the moment as user's cluster should always exist.

src/spark/cluster.py
Tianhao-Gu: 👍

src/spark/cluster.py
Tianhao-Gu: you mean when the method is being called? I think that's fine.

MrCreosote: e.g. from https://github.com/kbase/cdm-jupyterhub/pull/175/files#diff-5c1372251bcdee275758e2d3511fa8c50bd6c52a792739bd5d93d07fac46d6adR70

So when the notebook starts, IIRC, it'll print those two lines

MrCreosote (approved):

src/spark/cluster.py
Tianhao-Gu: 👍  I see. yea. moved print statements inside of `if not force`.

PR #190, opened 2025-05-19
add spark-redis package

Tianhao-Gu:


Dockerfile
```diff
@@ -35,6 +35,7 @@ ENV HADOOP_AWS_VER=3.3.4
 ENV DELTA_SPARK_VER=3.2.0
 ENV SCALA_VER=2.12
 ENV POSTGRES_JDBC_VER=42.2.23
+ENV SPARK_REDIS_VER=3.1.0
```
MrCreosote, line 38: Is there any reason not to build jupyterhub on top of cdm-spark-standalone? Then you don't need this

MrCreosote (approved):

Dockerfile
Tianhao-Gu: Yea. It’s in my todo list. It’s probably a good idea to remove all duplicates from both repo.

PR #220, opened 2025-07-28
apply block formatter

Tianhao-Gu:


src/jupyterhub_config/custom_kube_spawner.py
```diff
@@ -207,56 +232,82 @@ def _ensure_virtual_environment(self, user_env_dir: Path):
...
         # In KubeSpawner, environment is a dict where the value can only be a strong
         # ref: https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html#kubespawner.KubeSpawner.environment
-        self.environment.update({key: str(value) for key, value in os.environ.items() if key not in self.environment})
+        self.environment.update(
+            {
+                key: str(value)
+                for key, value in os.environ.items()
+                if key not in self.environment
+            }
+        )
```
MrCreosote, line 257: It's funny, with black formatting the brevity of list comprehensions is completely gone. A standard for loop would be fewer lines

MrCreosote (approved):

src/jupyterhub_config/custom_kube_spawner.py
Tianhao-Gu: That's true. It's not perfect. But I was thinking it could be useful if everyone is using the same formatter or linter. Maybe build something in GHA. Haven't completely thought this through.

PR #223, opened 2025-07-29
add governance endpoints

Tianhao-Gu:


src/minio_governance/models.py
```diff
@@ -0,0 +1,122 @@
...
+
+from pydantic import BaseModel
+
+
+class CredentialsResponse(BaseModel):
+    """Response model for user credentials"""
+
+    username: Annotated[str, "Username"]
+    access_key: Annotated[str, "MinIO access key (same as username)"]
+    secret_key: Annotated[str, "MinIO secret key (fresh on each request)"]
```
MrCreosote, line 16: How does this documentation show up in the notebook?

src/minio_governance/models.py
```diff
@@ -0,0 +1,122 @@
...
+    user_home_policy: Annotated[PolicyInfo, "User's home policy"]
+    user_system_policy: Annotated[PolicyInfo, "User's system policy"]
+    group_policies: Annotated[List[PolicyInfo], "Policies from group memberships"] = []
+    total_policies: Annotated[int, "Number of active policies"]
+
+
+class HealthResponse(BaseModel):
+    """Response model for health check"""
+
+    status: Annotated[str, "Service status"]
```
MrCreosote, line 93: This seems like it could use some more info so the user knows what to expect

src/minio_governance/models.py
```diff
@@ -0,0 +1,122 @@
...
+
+    username: Annotated[str, "Username"]
+    access_key: Annotated[str, "MinIO access key (same as username)"]
+    secret_key: Annotated[str, "MinIO secret key (fresh on each request)"]
+
+
+class SharePathRequest(BaseModel):
+    """Request model for sharing a path"""
+
+    path: Annotated[str, "S3 path to share"]
```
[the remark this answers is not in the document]
Tianhao-Gu: yes. I don't plan to expose those endpoints directly to users. Instead, I'll provide a simpler interface, such as a share(namespace, table, permission) function, to make usage easier.

src/minio_governance/models.py
Tianhao-Gu: hmmm I am not sure I am following you. It would be something like a dict.

src/minio_governance/models.py
Tianhao-Gu: I don't think users will call healthcheck.

src/minio_governance/models.py
MrCreosote: This is moot since these classes aren't intended for user consumption if I understand you correctly

src/minio_governance/models.py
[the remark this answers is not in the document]
MrCreosote: Ok, I wouldn't hurt to document that these functions and models are for internal consumption vs users, as I assumed the latter

src/minio_governance/models.py
Tianhao-Gu: no - I will do another set of helper functions for users to call. So that they don't need to worry about building request model, getting client, etc.

src/minio_governance/client.py
```diff
@@ -81,3 +95,73 @@ def _handle_response(self, response: httpx.Response) -> Any:
                 error_type=error_type,
                 message=error_message,
             )
+
+    def health_check(self) -> HealthResponse:
+        """Check service health"""
+        response = self._get_client().get("/health")
+        data = self._handle_response(response)
+        return HealthResponse(**data)
```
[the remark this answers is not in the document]
MrCreosote: :+1:

PR #225, opened 2025-07-30
update docker spawner to use the data governance minIO creds

Tianhao-Gu:


src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -229,24 +231,42 @@ def _configure_environment(self, user_dir: Path, user_env_dir: Path, username: s
...
             )
-            self.environment["MINIO_ACCESS_KEY"] = self.environment[
-                "MINIO_RW_ACCESS_KEY"
-            ]
-            self.environment["MINIO_SECRET_KEY"] = self.environment[
-                "MINIO_RW_SECRET_KEY"
-            ]
-            # USAGE_MODE is used by the setup.sh script to determine the appropriate configuration for the user.
-            self.environment["USAGE_MODE"] = "dev"
+        client = DataGovernanceClient(kbase_token=kbase_token)
```
MrCreosote, line 241: How does the url get set here? env var?

src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -229,24 +231,42 @@ def _configure_environment(self, user_dir: Path, user_env_dir: Path, username: s
...
+        # Get credentials for the current user - it will also create a new user if it doesn't exist
+        credentials = client.get_credentials()
+        self.log.info(
+            f"Retrieved governance credentials for user {username}: access_key='{credentials.access_key}'"
+        )
+
+        if self.environment.get("USE_DATA_GOVERNANCE_CREDENTIALS", "false") == "true":
+            # Set user-specific MinIO credentials
+            self.environment["MINIO_ACCESS_KEY"] = credentials.access_key
+            self.environment["MINIO_SECRET_KEY"] = credentials.secret_key
```
MrCreosote, line 252: I assume the user will need to get these creds to log in to the minio UI. How do they do that?

MrCreosote, line 252: If a user accidentally leaks their creds, is there a way to remove their old creds and get a new set?

src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -229,24 +231,42 @@ def _configure_environment(self, user_dir: Path, user_env_dir: Path, username: s
...
 
-        if self._is_rw_minio_user():
-            self.log.info(
-                f"MinIO read/write user detected: {self.user.name}. Setting up minio_rw credentials."
+        # Get user-specific MinIO credentials from governance service
+        # Check if KBASE_AUTH_TOKEN is available (should be set by pre_spawn_start)
+        kbase_token = self.environment.get("KBASE_AUTH_TOKEN")
+        if not kbase_token:
+            raise ValueError(
+                "KBASE_AUTH_TOKEN not found in spawner environment. Authentication required for governance client."
```
MrCreosote, line 239: Should this and the other credential fetching stuff be inside the if block below, so that this error occurs only if use_data_gov_creds is true?

src/jupyterhub_config/custom_docker_spawner.py
```diff
@@ -229,24 +231,42 @@ def _configure_environment(self, user_dir: Path, user_env_dir: Path, username: s
...
+                ]
+                self.environment["MINIO_SECRET_KEY"] = self.environment[
+                    "MINIO_RW_SECRET_KEY"
+                ]
+            else:
+                self.log.info(
+                    f"Non-admin user detected: {self.user.name}. Removing admin credentials."
+                )
+                self.environment.pop("MINIO_RW_ACCESS_KEY", None)
+                self.environment.pop("MINIO_RW_SECRET_KEY", None)
```
MrCreosote, line 269: I assume if things get to this point, then `MINIO_ACCESS_[KEY|SECRET]` were already set in code not shown in this PR, so the user defaults to those creds?

src/jupyterhub_config/custom_docker_spawner.py
Tianhao-Gu: currently they do `os.environ['MINIO_ACCESS_KEY']` accordingly to the [guide](https://github.com/kbase/cdm-jupyterhub/blob/main/docs/dev_guide.md#readwrite-minio-username-and-password). I'm planning to implement a magic command to make it more convenient for users.

Tianhao-Gu: We do not store users' MinIO credentials. Instead, new credentials are generated each time a user logs in. Users can also manually rotate their credentials by calling `get_minio_credentials`, which is the same method invoked during the login process.

src/jupyterhub_config/custom_docker_spawner.py
Tianhao-Gu: The self.environment values from spawner will all be set to user's os env var - maybe that's what you are asking?

src/jupyterhub_config/custom_docker_spawner.py
MrCreosote: :+1:

src/jupyterhub_config/custom_docker_spawner.py
MrCreosote: Ok, the error message is slightly misleading in that case, but not a huge deal as this shouldn't happen IIUC

src/jupyterhub_config/custom_docker_spawner.py
MrCreosote: yeah, I'm assuming they'll get the creds set in the config map for the spawner, right?

Once the governace is live will those be removed?

Tianhao-Gu: yea.

src/jupyterhub_config/custom_docker_spawner.py
Tianhao-Gu: Saves us a lot of maintain and development cycle. I didn't expect users will need to login to minIO every time. It's also in the [design doc](https://github.com/kbase/cdm-minio-service/blob/main/docs/design.md) - no one ever raised any concern.

MrCreosote: Looks like no one ever reviewed it :/

src/jupyterhub_config/custom_docker_spawner.py
MrCreosote: It's talking about the governance client makes it sound like user_data_gov_creds is on. Like I said, minor

MrCreosote (approved):

PR #227, opened 2025-08-01
update spark session to use user's warehouse

Tianhao-Gu:


src/spark/utils.py
```diff
@@ -79,8 +77,9 @@ def _get_s3_conf() -> Dict[str, str]:
     Returns:
         Dictionary of S3/MinIO Spark configuration properties
     """
-    # TODO: Get warehouse directory from data governance service
-    warehouse_dir = DEFAULT_DELTALAKE_WAREHOUSE_DIR
+    # Set warehouse directory to the user's SQL warehouse
+    governance_client = DataGovernanceClient()
```
MrCreosote, line 81: How come in the previous PR the token had to be set explicitly but here it doesn't?

Tianhao-Gu: Correct. In the Docker/Kubernetes spawner setup, code execution happens on the hub server end. I pass each user's token through the spawner's `environment` variables. However, for utility functions, we expect the code to run within the user's notebook environment, where all necessary OS environment variables should already be set by the spawner.

MrCreosote: :+1:

MrCreosote (approved):

PR #243, opened 2025-08-27
support users to write to tenant sql warehouse

Tianhao-Gu:


docs/tenant_sql_warehouse_guide.md
```diff
@@ -0,0 +1,115 @@
...
+spark = get_spark_session(tenant_name="kbase")
+create_namespace_if_not_exists(spark, "research_data", append_target=True)  # Creates "kbase_research_data"
+```
+
+### Explicit Control
+```python
+# Explicitly disable prefixing (same as default)
+create_namespace_if_not_exists(spark, "global_reference", append_target=False)
+# Creates "global_reference" with no prefix
+```
```
MrCreosote, line 115: Could you add an example for how to switch from one warehouse to another? IIUC I don't think you can just call get_spark_session again or am I wrong?

src/db_ops/spark_db_utils.py
```diff
@@ -8,17 +8,43 @@
...
+    :param append_target: If True, prepends target name from warehouse directory
+                         (e.g., "john_default" or "research_team_experiments").
+                         If False, uses namespace as-is.
     :return: None
     """
-
+    try:
+        # Extract user/tenant name from warehouse directory if append_target is enabled
+        if append_target:
+            warehouse_dir = spark.conf.get('spark.sql.warehouse.dir', '')
```
MrCreosote, line 31: When would this not be set?

src/spark/utils.py
```diff
@@ -242,25 +259,36 @@ def get_spark_session(
...
         >>> # With custom scheduler pool
         >>> spark = get_spark_session("MyApp", scheduler_pool="highPriority")
 
         >>> # Local development
         >>> spark = get_spark_session("TestApp", local=True)
     """
     # Generate app name if not provided
+    current_spark = SparkSession.getActiveSession()
+    if current_spark:
+        current_spark.stop()
```
MrCreosote, line 290: Hm, are there any drawbacks to this? I assume there's some reason why we didn't do it before

src/db_ops/spark_db_utils.py
Tianhao-Gu: Users technically can create their own Spark session, but this is highly unlikely. It’s here as a safeguard.

docs/tenant_sql_warehouse_guide.md
Tianhao-Gu: I added the stop so that users can just call get_spark_session without worrying about calling stop explicitly.

src/db_ops/spark_db_utils.py
MrCreosote: :+1:

docs/tenant_sql_warehouse_guide.md
MrCreosote: :+1:

src/spark/utils.py
Tianhao-Gu: That’s fine—the only situation where you’d switch from personal to tenant mode is when creating a tenant table, which requires stopping the current Spark session.

MrCreosote (approved):

MrCreosote (approved):

MrCreosote:
A PR so nice I approved it twice

PR #286, opened 2025-11-18
Update MFA status casing to match auth2 API

dauglyon:
Updates MFA status values from uppercase to match auth2 API changes.

USED -> Used, UNKNOWN -> Unknown (NotUsed was not present in code)

Ref: https://github.com/kbase/auth2/pull/502


src/jupyterhub_config/kb_jupyterhub_auth.py
```diff
@@ -88,11 +88,11 @@ async def authenticate(self, handler, data=None) -> dict:
                 ),
             )
 
-        # Validate MFA requirement - only allow USED status
-        if kb_user.mfa_status != "USED":
+        # Validate MFA requirement - only allow Used status
+        if kb_user.mfa_status != "Used":
```
MrCreosote: Personally I would just run .lower() on mfa_status but this is fine

src/jupyterhub_config/kb_jupyterhub_auth.py
MrCreosote, line 1: The builds need to pass otherwise there's no image to deploy. I"m guessing there needs to be a uv sync?

MrCreosote (changes requested):
@Tianhao-Gu do you want to review this?

MrCreosote:
The new review UI seems buggy, comments disappear

Tianhao-Gu (approved):
The code changes looks good - I also fixed the uv sync issue - that was on me!

MrCreosote (approved):
LGTM, thanks @Tianhao-Gu

MrCreosote:
Actually thinking about this - if you .lower() the MFA status then it's compatible with the current auth as well as new auth so we could just deploy it right now. If we deploy this cdmhub will break

